PR #3, opened 2018-02-04
geoip alias and intro tweak

mimugmail:


source/manual/aliases.rst
```diff
@@ -66,6 +68,17 @@ URL tables can be used to fetch a list of ip addresses from a remote server.
 There are several IP lists available for free, most notably are the "Don't Route
 Or Peer" lists from Spamhaus.
 
+-----
+GeoIP
+-----
+With GeoIP alias you can select one or more countries or whole continents to block
+or allow. Use the *toggle all* checkbox to select all countries within the given
+region. This feature was reworked with 17.7.7 and supersedes the GeoIP blocking via
```
fichtner: New paragraph for historical annotation. Nice touch here, that's the forward-compatible way to document this. :)

source/manual/aliases.rst
```diff
@@ -22,6 +22,8 @@ OPNsense offers the following alias types:
 +------------+------------------------------------------------------+
 | URL Tables | A table of ip addresses that can be fetched          |
 +------------+------------------------------------------------------+
+| GeoIP      | Select countries or whole continents                 |
```
fichtner: Stick to "regions", we don't support the ISO Continent Codes because we have no mapping ATM

source/intro.rst
```diff
@@ -71,7 +71,7 @@ OPNsense Core Features
   - Road warrior
   - IPsec
   - OpenVPN
-  - Legacy PPTP support
+  - Legacy PPTP support (via plugins)
```
fichtner: On second thought, let's remove this line completely

mimugmail:
Done .. what do you mean with paragraph thing?

source/manual/aliases.rst
fichtner: Break paragraph for last sentence here ^^^

mimugmail: ok now? :)

fichtner:
Yes, thank you!

PR #8, opened 2018-02-18
Two Factor rework

fabianfrz:
Please note that the app has not language setting.
The existing screenshots on that page are in dutch so I think it is not a huge issue.


AdSchellevis:
@fabianfrz looks good, thanks!

PR #14, opened 2018-03-09
add nextcloud docs

fabianfrz:
should await opnsense/core#2251


fabianfrz:
@fichtner when this is merged, can you render it and upload it to the docs server. There are few changes pending, which are not included on the docs server yet.

fichtner:
Merged, thank you!

fabianfrz:
@fichtner havn't expected that to be merged now because @AdSchellevis is waiting for something in core:
opnsense/core#2289

BTW: I like that the UI has not changed due to this transformation of the code in the backend.

fichtner:
It’s ok if the docs are a little ahead of he release.

Yes, I love the backup framework even though the fronted is legacy at the moment.  But we already discussed to add a controller and view to render each one in a separate page that could land in the menu instead of the current page.

PR #20, opened 2018-03-31
fix masks in shaper howto

mimugmail:
fixes core/#1542


fichtner:
Merged, many thanks for this one!

PR #21, opened 2018-04-02
fix shaper heading; add note about Android

fabianfrz:
screenshot_20180402_112325


fabianfrz:
@ruggerio wrote that it is ok. So it is ready.

fichtner:
Merged, thank you!

PR #26, opened 2018-06-20
www/nginx: add a basic load balancer how to

fabianfrz:
add a short how to for load balancing with nginx.


fichtner:
Merged, thanks!

PR #35, opened 2018-07-21
Outbount NAT for VPN client

kalos:
Add a note that explain how to allow the VPN server network to connect to the VPN client network.


fabianfrz:
you don't need NAT. You can configure routing as well (for example static routes)

kalos:
Ok, please leave a note to explain that in the documentation.
I think it would be useful.
thanks!

AdSchellevis:
I'm not sure what we should add here, if NAT is required has more todo with your network design then the technology used (physical interfaces., vpn, ..).
If someone wants to create some basic networking documentation that would probably be helpful, I just don't think we should clutter technology with network design choices.

PR #36, opened 2018-07-31
CARP warning

mimugmail:
Make a clear warning that mixed interfaces will break CARP.

Reference:

https://www.reddit.com/r/OPNsenseFirewall/comments/92usm2/need_help_with_carp_setup_on_vmware_esxi/e3a71sp


AdSchellevis:
@mimugmail it shouldn't be required to have the exact same drivers (the carp traffic doesn't know anything about the driver), although I can imagine driver issues in some scenario's. I would change the warning to a Tip about compatibility issues.

mimugmail:
Urgs .. I create a new one when Wireguard is merged. I screwed up my docs-fork :/
Drivers I also only read about, but Assignment has to be the same, so Warning for sure ...

PR #37, opened 2018-08-17
Add WireGuard

mimugmail:


source/manual/how-tos/wireguard-s2s
```diff
@@ -0,0 +1,63 @@
+============================
+WireGuard Site-to-Site Setup
+============================
+
+------------
+Introduction
+------------
+
+WireGuard is an extremely simple yet fast and modern VPN that utilizes state-of-the-art cryptography.
```
fichtner: minor concern not needed to be fixed now: marketing boilerplate should not be in technical documentation :D

source/manual/how-tos/wireguard-s2s
```diff
@@ -0,0 +1,63 @@
...
+It aims to be faster, simpler, leaner, and more useful than IPSec, while avoiding the massive headache.
+It intends to be considerably more performant than OpenVPN. WireGuard is designed as a general purpose
+VPN for running on embedded interfaces and super computers alike, fit for many different circumstances.
+Initially released for the Linux kernel, it is now cross-platform and widely deployable. It is currently
+under heavy development, but already it might be regarded as the most secure, easiest to use, and simplest
```
fichtner: most secure, hold that thought until the first security problems are reported and the linux kernel is pwned because of it ;)

mimugmail:
True that .. the introduction is very aggressive. :)

fichtner:
1.0 in 18.7.3 ? ;)

mimugmail:
No, ATM I'm building 0.4 with a tab for showing live config without keys.
I have a feeling that there'll come some more questions and this would help to troubleshoot :)

fichtner:
ok, sure, no rush

maybe we want to pull this in anyway and mark it "development only" ?

mimugmail:
This would be very cool since I also want to push my AzireVPN docs and client/server internet breakout docs to the official spaces before released.

source/manual/how-tos/wireguard-s2s
```diff
@@ -0,0 +1,59 @@
...
+
+WireGuard is a simple and fast modern VPN. It aims to be faster and simpler than IPSec. It intends to be
+considerably more performant than OpenVPN. Initially released for the Linux kernel, it is now cross-platform
+and widely deployable. It is currently under heavy development.
+
+---------------------
+Step 1 - Installation
+---------------------
+
+Go to **System->Firmware->Plugins->** and search for **os-wireguard**.
```
fichtner: this should read "os-wireguard-devel" then and before step 1 should be a warning box saying it's only for development mode.

mimugmail:
Should be better now ...

fichtner:
Perfect... merged, thanks!

mimugmail:
When is PR deadline for 18.7.3?

fichtner:
Monday. Since FreeBSD issued new SAs the release was pushed back. I originally planned for yesterday.

PR #41, opened 2018-09-04
Lan Bridge How To

marjohn56:
Instructions on how to setup a LAN Bridge


source/manual/how-tos/lan_bridge.rst
```diff
@@ -0,0 +1,90 @@
+**How to set up a LAN Bridge**
+========================
+
+**Original Author:** Martin Wasley
```
fabianfrz: is that needed?

It is usually not in the docs but visible in the git logs…

marjohn56:
Which bit.. the originally authored bit or the fact that the underline was too short, I corrected that after I loaded into my web server and saw the error. As for the authored by bit, when I didn't add it on the Zen doc, it was added for me.. :)

source/manual/how-tos/lan_bridge.rst
fichtner: it's ok to put this in, some things can be imported from other articles or so and the docs are for everyone to share

fichtner: (feel free to update your own)

marjohn56: I'm quite happy not to have it, or I could use an alias.. 😋

fichtner:
And merged, thanks.

PR #45, opened 2018-09-17
Update carp.rst

mimugmail:
As discussed with @AdSchellevis switched Warning to Note

### DIFF
--- a/source/manual/how-tos/carp.rst
+++ b/source/manual/how-tos/carp.rst
@@ -66,6 +66,13 @@ can be found under System -> High Availability.
 Setup interfaces & basic firewall rules
 -----------------------------------------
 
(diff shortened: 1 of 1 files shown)


AdSchellevis:
@mimugmail the driver itself don't need to be the same, it's convenient to use the same hardware, mostly for troubleshooting reasons, but it can function with different drivers as well (em0 on master vs igb0 on the backup should work without issues). Can we change the `Make sure you are using` for something like `It's advisable to use`?

mimugmail:
I removed the driver section and only left the real critical issue, changed to warning as it won't work with mixed optX assignments.

AdSchellevis:
@mimugmail thanks!

PR #47, opened 2018-09-19
Add tunable warning to dynamic_routing.rst

mimugmail:
Predicted it's in 18.7.5 :D

### DIFF
--- a/source/manual/dynamic_routing.rst
+++ b/source/manual/dynamic_routing.rst
@@ -3,6 +3,12 @@ Dynamic Routing
 ===============
 
 .. Warning::
+    With OPNsense version 18.7.5 the frr package was updated to version 5. It's strongly advised to higher
+    the kern.ipc.maxsockbuf value via **Tunables**. Go to **System->Settings->Tunables** and check if there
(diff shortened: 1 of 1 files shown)


source/manual/dynamic_routing.rst
```diff
@@ -2,6 +2,12 @@
 Dynamic Routing
 ===============
 
+.. Warning::
+    With OPNsense version 18.7.5 the frr package was updated to version 5. It's strongly advised to higher
```
fichtner: I'd propose something like:

OPNsense 18.7 uses FRR version 5.

No need to be picky. The advise given is also ok for FRR < 5 so no harm done. :)

source/manual/dynamic_routing.rst
```diff
@@ -3,7 +3,7 @@ Dynamic Routing
 ===============
 
 .. Warning::
-    With OPNsense version 18.7.5 the frr package was updated to version 5. It's strongly advised to higher
+    With OPNsense version 18.7 the frr package was updated to version 5. It's strongly advised to higher
```
fichtner: sorry, two more: frr -> FRR, higher -> increase

fichtner:
Merged, thanks!

PR #48, opened 2018-09-20
new api docs

mimugmail:
- wireguard
- netsnmp
- ntopng
- vnstat

### DIFF
--- a/source/development/api/plugins/netsnmp.rst
+++ b/source/development/api/plugins/netsnmp.rst
@@ -1,0 +1,29 @@
+netsnmp
+~~~~~~~
+
+.. csv-table:: Settings
+   :header: "Method", "Module", "Controller", "Command", "Parameters"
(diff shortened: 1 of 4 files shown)


AdSchellevis:
@mimugmail thanks! Just out of curiosity, have you used https://github.com/fvanroie/opnsense_MakeDocs to generate these?  In the future a python script would be more practical for us to integrate in our regular toolchain, but the powershell script should deliver the correct output.

mimugmail:
No wasn't aware of :) Just search and replace ...

AdSchellevis:
no problem, as long as the naming is correct, we can always replace it with an automated solution in our release engineering process.

fvanroie:
What is the best trigger for updating the docs? Should it be updated on regular intervals or with each point release?

AdSchellevis:
we try to update when there's new content, but eventually we should look into a more solid release process (also since development and production can differ).

PR #50, opened 2018-09-25
Add bind and frr to API

mimugmail:
If I'm correct only nginx is missing, but there's too much development to include now.


AdSchellevis:
@mimugmail thanks!

PR #54, opened 2018-10-01
Update ipsec-road.rst

mimugmail:
Added a note after save. Make a difference when IPsec was started before or not.

### DIFF
--- a/source/manual/how-tos/ipsec-road.rst
+++ b/source/manual/how-tos/ipsec-road.rst
@@ -238,6 +238,12 @@ And Apply changes:
 
 -----------------------------
 
+.. Note::
+
+   If you already had IPsec enabled and added Road Warrior setup, it's important to 
+   restart the whole service via **System->Diagnostics->Services->Strogswan** since applying
+   configuration only reloads it, but a restart also loads the required modules of strongswan.
+
 ------------------------
 Step 4 - Add IPsec Users
 ------------------------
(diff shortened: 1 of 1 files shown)


source/manual/how-tos/ipsec-road.rst
```diff
@@ -238,6 +238,12 @@ And Apply changes:
 
 -----------------------------
 
+.. Note::
+
+   If you already had IPsec enabled and added Road Warrior setup, it's important to 
+   restart the whole service via **System->Diagnostics->Services->Strogswan** since applying
```
fichtner: Doesn't the service widget do this on all the IPsec pages as well?

fichtner: But the reload button would so the same thing on dashboard, diagnostics and page control unless there is a bug...?

mimugmail: It needs a restart, not a relaod. A reload is done on every change. The IPSec pages don't have an own services widget. And the Services widget via Dashboard can also be removed by the user, that's why is I have choosen the most stable approach.

fichtner: Err, lol, I broke it in 2016 and nobody noticed. :) https://github.com/opnsense/core/commit/e99ae7ad17b

mimugmail: Dont hate the Player :) I'll change it tomorrow .. gn

fichtner: No, no, thanks for finding this! gn

fichtner:
Merged, thanks. I'll add the service control to 18.7.6.

mimugmail:
Thx, I'll rewrite the road warrior part today fitting more clients with the new multiselect for hash and DH

fichtner:
great, looking forward to merging it ❤️

PR #58, opened 2018-10-20
fix menu

mimugmail:
"===" Will add a new menu in rst as you can see in this Overview:
https://docs.opnsense.org/manual/howtos.html

Overview and Interface Assignment belongs to ZeroTier ...


fichtner:
Merged, thanks!

PR #73, opened 2018-11-01
Extend the cellular modem howto

ktk:
I've extended the cellular modem howto quite a bit, as I had troubles getting my to card work. Everything I figured out so far is represented in the changes and should help other users debugging their setup, in case not everything works out of the box.

Note that I did not write RST so far, maybe someone can check if my syntax is correct. Preview in github looks ok but I could for example not test the "info" boxes.

### DIFF
--- a/source/manual/how-tos/cellular.rst
+++ b/source/manual/how-tos/cellular.rst
@@ -168,6 +168,12 @@ providers required factory resets (for whatever reason) to get them to work prop
   As mentioned above, the SIM card needs to be available. See the hints
   in the first step of this tutorial to make sure everything is prepared properly.
 
+.. Note::
+  
+  Some Sierra Wireless modems still seem to need a specific init string to work
+  properly. One that seems to work for multiple users and LTE cards is ``&F0E1Q0 +CMEE=2``. In any case you should first try without init string and only give it
+  a try if you could not get any connection without. You can add this in **Interfaces->Point-to-Point->Devices->Your particular device->Advanced Options->Init String**.
+
 When the device seems to work properly then checkout if the interface was assigned
 an IP address, go to **Interfaces->Overview** and click on the WAN interface to
 see the details.
--- a/source/manual/how-tos/cellular.rst
+++ b/source/manual/how-tos/cellular.rst
@@ -71,7 +71,6 @@ Now, check if SIM card is available and can be accessed:
    cu -l /dev/cuaU0.2
    Connected
    AT+CPIN?
-   OK
    +CPIN: READY
 
 There are multiple things that can go wrong here, the SIM card might not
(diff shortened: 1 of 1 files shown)


fichtner:
@ktk thanks for the massive update. ❤️ 

handed over to @AdSchellevis as I don't know enough about this topic

ktk:
While we are at it, there are a few things that should IMHO be added to the guide but I did not figure them out yet:

- what code/component is doing the connection in OPNsense? I thought it's pppd but I think I never found a process
- How can I add APN configs for providers that are not in the list yet?
- Once the connection is up, how can I still send AT commands to the card to get stuff back like signal quality etc? See my question in the [forum](https://forum.opnsense.org/index.php?topic=9828.0)

fichtner:
1. FreeBSD package "mpd5", daemon "mpd"
2. https://github.com/opnsense/core/blob/master/contrib/mobile-broadband-provider-info/serviceproviders.xml
3. not sure

ktk:
@fichtner great tnx. Who is starting/triggering/configuring mdp?

Will do a pull request for my provider in Indonesia

fichtner:
https://github.com/opnsense/core/blob/5c6e9ff7e584b499a3bc73cac0c35f5118ef4f7f/src/etc/inc/interfaces.inc#L1163

The services file is no longer in sync with upstream if it still exists. I tried to sync it but had no good means to test the changes as their syntax changed slightly and impact on our code base was therefore  unclear...

AdSchellevis:
@ktk point [3] for Huawei it has a device (some rather old code in https://github.com/opnsense/core/blob/master/src/sbin/3gstats.php), but for Sierra wireless I've never seen that part working. What card type are you using, maybe there's some reference guide online.

Thanks for the documentation by the way, nice to see it improve like this.

AdSchellevis:
@ktk I'll check locally if the format is correct, text improvements look fine

AdSchellevis:
this should be your reference AT command guide https://www.sierrawireless.com/resources/documents/mp/2130309_mp7xx_at_command_reference.pdf 
technical specs seem to be available here https://source.sierrawireless.com/~/media/support_downloads/airprime/74xx/4116237_airprime_mc7430_product_technical_specification_r8.ashx

I'm not entirely sure how this should work with Sierra wireless, I had one of those cards a long time ago, but I'm not sure if it accepted connections on two ports or only one. It could also be that mpd should signal the status and you should grep it out of the mpd log for sierra cards.

ktk:
@AdSchellevis jep I have the command reference thanks, I will post on the FreeBSD list to figure out how I could interact with that thing. When I figure it out I will do an update to the guide again.

AdSchellevis:
@ktk thanks!

PR #72, opened 2018-10-31
rewrite IPsec remote access documentation

mimugmail:
This is my complete rewrite of RA doc.
For review it might be easier to view my branch. 

This is the starting point: 
https://github.com/mimugmail/docs/blob/matrix/source/manual/how-tos/ipsec-rw.rst

At the end it presents you a matrix, on the right for each mechanism one single page like:
https://github.com/mimugmail/docs/blob/matrix/source/manual/how-tos/ipsec-rw-srv-mschapv2.rst

For every client OS you can view the screenshots needed to get it running (if compatible) like:
https://github.com/mimugmail/docs/blob/matrix/source/manual/how-tos/ipsec-rw-android.rst

What's missing is iOS and mac since I get devices next week or so. It's also not linked yet so might be safe to merge.

Please take care and time for review :)

PR #78, opened 2018-11-07
Fix scaling of small images, fix typos, clarify IPsec s2s

MichaelDeciso:
This manual leaves IPsec Road Warrior alone, since there is already a PR open for that.
It _does_ touch IPsec S2S, which is not covered by https://github.com/opnsense/docs/pull/72.

### DIFF
--- a/source/manual/how-tos/ips-feodo.rst
+++ b/source/manual/how-tos/ips-feodo.rst
@@ -57,7 +57,7 @@ First apply the configuration by pressing the **Apply** button at the bottom of
 the form.
 
 .. image:: images/applybtn.png
-    :width: 100%
+    :scale: 100%
 
 ---------------
 Fetch Rule sets
@@ -70,7 +70,7 @@ To do so: select Enabled after each one.
 
 To download the rule sets press **Download & Update Rules**.
 
-.. image:: images/downloadbtn.png
+.. scale:: images/downloadbtn.png
     :width: 100%
 
 -----------------------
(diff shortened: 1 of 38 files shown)


source/manual/how-tos/ips-feodo.rst
```diff
@@ -70,7 +70,7 @@ To do so: select Enabled after each one.
 
 To download the rule sets press **Download & Update Rules**.
 
-.. image:: images/downloadbtn.png
+.. scale:: images/downloadbtn.png
```
fichtner: image -> scale, not width -> scale like the other two?

also :scale: was changed to :width: in earlier commits because it squashes larger images

MichaelDeciso: 1. Oops
2. This is why:

ipsec-s2s-een-beetje-groot

fichtner: hehe, ok. maybe there is a more portable setting than cranking every :scale: and :width: to 100% ?

MichaelDeciso: That could be the case, indeed. I was following @jschellevis ' instructions here.
For small images that don't need resizing, we might be able to leave out both instructions.

MichaelDeciso:
From the Sphinx manual:
```
scale : integer percentage (the "%" symbol is optional)

    The uniform scaling factor of the image. The default is "100 %", i.e. no scaling.
```
In other words: it can be left out. I discussed this with @jschellevis and he agreed on doing so.

fichtner:
nice, thanks

AdSchellevis:
@MichaelDeciso thanks!

PR #81, opened 2018-11-09
Remove documentation about GeoIP via IPS

MichaelDeciso:
I have tested building the docs. It builds successfully and the reference to the section works correctly.


source/manual/how-tos/ips-geoip.rst
```diff
@@ -2,108 +2,6 @@
 IPS GeoIP Blocking
 ==================
 
-This tutorial explains how to setup the IPS system to block IPs based on their
-geographic location. This option is made possible by the integration of the
-Maxmind GeoLite2 Country database. More information can be found here: http://dev.maxmind.com/geoip/geoip2/geolite2/
+This feature has been superseded by using aliases in 17.7, and has been removed in 19.1.
```
AdSchellevis: @MichaelDeciso technically it's scheduled for removal :)

MichaelDeciso: That's correct, indeed. I was trying to future-proof it a bit.

Would you like me to change it to say 'has been scheduled for removal in 19.1'?

AdSchellevis: maybe better to remove the "and has been.." part, superseded should be enough and is future proof.

PR #85, opened 2018-11-11
add nginx IP ACLs

fabianfrz:


source/manual/how-tos/nginx_ip_acl.rst
```diff
@@ -0,0 +1,62 @@
...
+
+Background Information
+======================
+
+IP based ACLs can be externally used to allow access (whitelist strategy) to a
+specific web service only by customers so you can easily get rid of most of the
+malicious traffic to the application server.
+This also has some downsides like the site is probably
+not visible to search engines and will not be indexed.
+On the other hand you can also blacklist (blacklist strategy) some bot IPs,
```
MichaelDeciso:
```suggestion
On the other hand you can also blacklist (blacklist strategy) some bot IPs and
```

source/manual/how-tos/nginx_ip_acl.rst
```diff
@@ -0,0 +1,62 @@
...
+    as the only protection for your webservices.
+
+
+Background Information
+======================
+
+IP based ACLs can be externally used to allow access (whitelist strategy) to a
+specific web service only by customers so you can easily get rid of most of the
+malicious traffic to the application server.
+This also has some downsides like the site is probably
```
MichaelDeciso:
```suggestion
This also has some downsides: for example, the site will probably
```

source/manual/how-tos/nginx_ip_acl.rst
```diff
@@ -0,0 +1,62 @@
...
+not visible to search engines and will not be indexed.
```
MichaelDeciso:
```suggestion
not be visible to search engines and will therefore not be indexed.
```

source/manual/how-tos/nginx_ip_acl.rst
```diff
@@ -0,0 +1,62 @@
...
+
+.. image:: images/nginx_ip_acl_01_list_view.png
+
+Click the + button to create a new ACL.
+
+.. image:: images/nginx_ip_acl_02_create_acl_view.png
+
+Next enter a reasonable title, for example here "Allow Private IPs" was used.
+Now the different IP addresses or IP ranges can be entered. In this case some
+common private IP ranges were allowd and the default rule was set to block.
```
MichaelDeciso:
```suggestion
common private IP ranges were allowed and the default rule was set to block.
```

source/manual/how-tos/nginx_ip_acl.rst
```diff
@@ -0,0 +1,62 @@
...
+A new line can be added by clicking the + icon while trash can deletes the row.
```
MichaelDeciso:
```suggestion
A new line can be added by clicking the + icon while the trash can icon deletes the row.
```

source/manual/how-tos/nginx_ip_acl.rst
```diff
@@ -0,0 +1,62 @@
...
+This means that this searvice should be only visible internally.
```
MichaelDeciso:
```suggestion
This means that this service should be only visible internally.
```

source/manual/how-tos/nginx_ip_acl.rst
```diff
@@ -0,0 +1,62 @@
...
+
+.. Warning::
+    Keep in mind that carrier grade NAT (CGN) may cause some trouble with this
```
MichaelDeciso:
```suggestion
    Keep in mind that carrier grade NAT (CGN) may cause some trouble with these
```

source/manual/how-tos/nginx_ip_acl.rst
```diff
@@ -0,0 +1,62 @@
...
+    ACLs too. Please check how your traffic is handled first.
+
+
+Assign it to a Location, HTTP- or Stream-Server
```
MichaelDeciso:
```suggestion
Assign it to a Location, HTTP or Stream-Server
```

source/manual/how-tos/nginx_ip_acl.rst
fabianfrz: after a column a capital should follow, otherwise I would accept your suggestion

fabianfrz:
```suggestion
This also has some downsides: For example, the site will probably
```

MichaelDeciso: No, it should not. Not in English, at least.

fabianfrz: https://www.grammarly.com/blog/capitalization-after-colons/

-> 3rd heading

fichtner:
Merged, thanks!

PR #90, opened 2018-12-06
Fix #89: Make user manager doc reflect current GUI

### LINKED ISSUE #89: Menu item "System > Access > Configuration" is gone
Referring to the documentation [here](https://wiki.opnsense.org/manual/users.html?highlight=radius), I am unable to find the menu entry for "System > Access > Configuration":
image

I was looking for the described settings in other sections for quite a while but did not succeed.
I was wondering whether it might be a bug?

MichaelDeciso:


AdSchellevis:
@MichaelDeciso thanks!

PR #96, opened 2018-12-20
Update bind.rst

Northguy:
Added a line to alter unbound settings to make advanced settings work

### DIFF
--- a/source/manual/how-tos/bind.rst
+++ b/source/manual/how-tos/bind.rst
@@ -97,3 +97,5 @@ you can set in **Services->Unbound DNS->General->Custom Options**.
     forward-zone:    
     name: „.“    
     forward-addr: 127.0.0.1@53530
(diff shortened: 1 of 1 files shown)


fichtner:
This is obsolete in 19.1 as we always bind localhost on 127.0.0.1 to avoid such issues to begin with. 😊

Northguy:
Does this mean that the dropdownbox will disappear in 19.1? As the seting of this dropdownbox was affecting my configuration.

fichtner:
No, drop-downs will stay there. Separate items like IP addresses or Localhost/Loopback will disappear as they are implied by the interface selected or now always on.

Northguy:
ok. clear.

Op ma 24 dec. 2018 om 14:33 schreef Franco Fichtner <
notifications@github.com>:

PR #98, opened 2018-12-29
www/nginx: add TLS fingerprint detection

fabianfrz:


fichtner:
Looks ok in general, but the freehand markup in the picture should be improved...

fabianfrz:
ok, I will try to recreate it - maybe with another tool - Maybe freehand drawing is not the right one here.

fabianfrz:
better?

fichtner:
Indeed, thanks! Merged.

PR #100, opened 2019-01-02
Update carp.rst

mimugmail:


AdSchellevis:
hmm, are you sure? I kind of expect it would work as long as you keep `Bind states to interface` to default.

mimugmail:
https://forum.netgate.com/topic/76845/pfsync-between-different-network-cards

mimugmail:
If I have 10 interfaces, one of them dedicated for pfsync and a state Change on Interface dmz2 should be triggered, what else other than the physical interface name could be used?

AdSchellevis:
I don't have the time to check it now, but I expect the `state-policy` is honoured on sync.... which isn't if-bound by default.

mimugmail:
No rush, If you have concerns, I'll double check :)

AdSchellevis:
It seems to be the case indeed, I think you can also rename your interfaces  (`ifconfig em3 name igb3`) to make sure the correct ifname is send with the pfsync packet, but we don't support that (too tricky).

https://github.com/opnsense/src/blob/ea449cda0d6a5c112080289a8357f10046de1ddc/sys/net/pfvar.h#L791

PR #101, opened 2019-01-03
Fix typo in carp.rst

MichaelDeciso:
172.18.0.0/24 is used throughout the guide, so this instance is a typo.


fichtner:
Nice catch, thanks!

PR #102, opened 2019-01-04
Added verification instructions to installation manual on wiki

arrogant2:
In reference to [this](https://github.com/opnsense/changelog/issues/3) issue on the changelog repo, here is my pull request.

I have added detailed instructions on how to verify the image. A few notes:

1. These instructions should be linked to from the download page, so they are easy to find.
2. I would recommend simply distributing the base64 sig file, rather than requiring the extra step. This will be less confusing.

I also recommend linking to this page from all future release announcements.


fichtner:
Merged, thanks!

PR #114, opened 2019-01-24
Create page about using certificates

MichaelDeciso:
I think there must be more to write about, but I'm not sure which direction to take. I do think this is at least a good start.

### DIFF
--- a/source/manual/certificates.rst
+++ b/source/manual/certificates.rst
@@ -2,7 +2,7 @@
 Using certificates
 ==================
 
-In OPNsense, certificates are used for encrypting traffic. To make using them easier, OPNsense allows creating
+In OPNsense, certificates are used for ensuring trust between peers. To make using them easier, OPNsense allows creating
 certificates from the front-end. In addition to that, it also allows creating certificates for other purposes,
 avoiding the need to use the ``openssl`` command line tool. Certificates in OPNsense can be managed from
 **System->Trust->Certificates**.


source/manual/certificates.rst
```diff
@@ -0,0 +1,18 @@
+==================
+Using certificates
+==================
+
+In OPNsense, certificates are used for encrypting traffic. To make using them easier, OPNsense allows creating
```
fabianfrz:
```suggestion
In OPNsense, certificates are used for ensuring trust between peers. To make using them easier, OPNsense allows creating
```

source/manual/certificates.rst
```diff
@@ -0,0 +1,18 @@
...
+certificates from the front-end. In addition to that, it also allows creating certificates for other purposes,
+avoiding the need to use the ``openssl`` command line tool. Certificates in OPNsense can be managed from
+**System->Trust->Certificates**.
+
+The following types of certificate can be generated in OPNsense:
+
+* Client
+* Server
+* Combined Client/Server
+* Certificate Authority
```
fabianfrz: they should also be explained (especially if you configure a server with a client certificate, the connection should be rejected -> especially important for the web interface)

source/manual/certificates.rst
```diff
@@ -0,0 +1,18 @@
...
+
+In addition to this, OPNsense can generate a Certificate Signing Request (CSR). This can be used if you want to create a
+certficate signed by an external CA.
```
fabianfrz: I would mention how they are used, i. e. https://github.com/opnsense/docs/blob/master/source/manual/how-tos/nginx_tls_auth.rst for client certificates.

source/manual/certificates.rst
MichaelDeciso: Hm, not sure what the best phrasing is here. @mimugmail do you have an idea?

mimugmail:
```suggestion
In OPNsense, certificate management is easy, as it allows creating
```

mimugmail: I'd make is more general as is suggestion :)

source/manual/certificates.rst
fichtner: From core perspective OpenVPN and IPsec.

source/manual/certificates.rst
MichaelDeciso: I have added a warning.

source/manual/certificates.rst
MichaelDeciso: OK, I listed those two. Normally, I'd put 3-4 examples, are there more or should I just leave it at two?

fichtner: Captive Portal and web proxy then. While the admin GUI uses it too it should stay on topic for user-facing parts, not admin parts.

source/manual/certificates.rst
fabianfrz: @MichaelDeciso at least not "encrypting traffic" as this is done by a cipher and never by a certificate.

source/manual/certificates.rst
fabianfrz: @MichaelDeciso perfect. Well written :+1:

source/manual/certificates.rst
MichaelDeciso: In the end, I just went with your suggestion.

PR #120, opened 2019-01-28
Write page about serial access

MichaelDeciso:


source/manual/how-tos/serial_access.rst
```diff
@@ -0,0 +1,28 @@
+=============
+Serial Access
+=============
+
+Besides the web frontend, SSH and a locally connected monitor (if your device supports it), OPNsense can also be
+controlled via serial. Accessing OPNsense via serial is similar to accessing via SSH, but unlike SSH, the system can
```
mimugmail:
```suggestion
controlled via serial. Accessing OPNsense via serial is similar to accessing via SSH, but unlike SSH, the system can be
```

source/manual/how-tos/serial_access.rst
```diff
@@ -0,0 +1,28 @@
...
+--------------------------------
+
+On Unix-like systems, you can connect to the serial console using the ``screen`` program, with a baud rate of 115200.
+The device name can differ per system and per serial device. For example, on the Deciso DEC630, accessed from macOS,
+the device is named ``/dev/tty.usbmodem1112421``. Entering the serial console thus involves opening a terminal and
+executing the following instruction:
+
+::
+
+  screen /dev/tty.usbmodem1112421 115200
```
fabianfrz: I use minicom with a profile

Config file:

```
cat /etc/minirc.opnsense 
# Machinell erzeugte Datei - Verwenden Sie "minicom -s" zum Ändern
pu port             /dev/ttyUSB0
```

so I just need to call:

```sh
sudo minicom opnsense
```

source/manual/how-tos/serial_access.rst
```diff
@@ -0,0 +1,28 @@
...
+
+If OPNsense is running, you will now be asked for your username and password. The credentials are the same as those
+used for SSH.
+
+A thing to note is that, unlike SSH, you will not be able to scroll (but you can pipe the output through a pager like
```
fabianfrz: I can scroll when the application does not use some thermial ui like vim in minicom.

fabianfrz:
Forgot that first: the configuration page in the web interface must be set correctly to make RS232 work.

source/manual/how-tos/serial_access.rst
MichaelDeciso: Possibly, but that doesn't work in `screen`.

source/manual/how-tos/serial_access.rst
```diff
@@ -0,0 +1,34 @@
...
+installation that was done via serial), serial access is already available.
+
+On Unix-like systems, you can connect to the serial console using the ``screen`` program, with a baud rate of 115200.
+The device name can differ per system and per serial device. For example, on the Deciso DEC630, accessed from macOS,
+the device is named ``/dev/tty.usbmodem1112421``. Entering the serial console thus involves opening a terminal and
+executing the following instruction:
+
+::
+
+  screen /dev/tty.usbmodem1112421 115200
```
AdSchellevis: it would be good to add common device names in other systems,

linux: /dev/ttyS0 or /dev/ttyUSB0
FreeBSD:  /dev/cuau0 or /dev/cuaU0
Windows usually uses COM[x]

source/manual/how-tos/serial_access.rst
AdSchellevis: better to remove, depends on settings.

source/manual/how-tos/serial_access.rst
fabianfrz: @AdSchellevis: "S" is usually a PCIe card or on the mainboard itself, while "USB" is for USB adapters. Maybe that should be noted as well.

You just need to know which one is the right device node because for example my dive computer interface is also a ttyUSB + number device. Maybe we should recommend reading `dmesg` when using USB devices?

MichaelDeciso: I have added a list of device name examples.

source/manual/how-tos/serial_access.rst
MichaelDeciso: I have rephrased it, since I think it's still an important gotcha, but I'll remove it if you think it's still confusing or unnecessary.

PR #123, opened 2019-01-30
monit: fix typo, format syslog keyword

fbrendel:
The _stop_ command should be _stop_ :smiley:


AdSchellevis:
@fbrendel nice catch!

PR #127, opened 2019-02-01
Write page on DHCP

MichaelDeciso:


fabianfrz:
This page should at least cover DHCP for IPv4 in a way most home users work with it. The current page does not really help. The others would be nice to have. 

I already wrote some minimal information regarding DHCP:

https://github.com/opnsense/docs/blob/master/source/manual/how-tos/pac.rst#wpad

source/manual/dhcp.rst
```diff
@@ -0,0 +1,43 @@
...
+* **Leases**: Shows all IP addresses that are handed out to clients (can be filtered to only show active and static leases).
+
+------------
+Using DHCPv4
+------------
+
+A typical DHCPv4 usage scenario is using it on your LAN with an IP range of 192.168.1.x, where x can be a number from 1
+through 254. This means a subnet mask of 255.255.255.0. The range can also be written as 192.168.1.0/24. (The “1” in
+the third group can also be another number, and there are also other ranges available for internal usage, such as
+10.0.0.0/24.)
```
fabianfrz:
```suggestion
10.0.0.0/8.)
```

just be correct

MichaelDeciso: Yes, the whole 10.0.0.0/8 subnet is reserved for this purpose, but what I'm talking about is what range you would set for a LAN network.

You would typically set 10.0.0.0/24 as a subnet for a LAN network, in the same way that you would set 192.168.76.0/24 or 10.211.55.0/24.

fabianfrz: How about just mention RFC1918 and link it to its document: https://tools.ietf.org/html/rfc1918#section-3

text change: "as defined in RFC1918"

PR #129, opened 2019-02-03
add changelogs

fabianfrz:
a short article about changelogs in OPNsense


AdSchellevis:
@fabianfrz looks good, thanks!

PR #130, opened 2019-02-05
sslvpn_client: Fix typos, update screenshots and interface descriptions

MichaelDeciso:


source/manual/how-tos/sslvpn_client.rst
```diff
@@ -188,8 +188,8 @@ For the first step we enter:
  **Username**       *Donald*
  **Password** (2x)  *S3cr3tP@ssw0rd*
  **Full name**      *Donald Duck*
- **Certificate**    *True*
- **OTP seed**       *True*
+ **Certificate**    *Check “Click to create a user certificate”*
+ **OTP seed**       *Check “Generate new secret”*
 ================== ===================
```
fabianfrz: now the border is too short

source/manual/how-tos/sslvpn_client.rst
```diff
@@ -290,7 +290,7 @@ Step 2 - Firewall Rules
 To allow SSL VPN client connections, we should allow access to the OpenVPN server
 port on the WAN interface. When using multiple servers we need to open up each port.
 
-For our configuration we only use one server accessible on udp port 1194.
+For our configuration we only use one server, accessible on UDP port 1194.
 
 .. image:: images/sslvpn_wan_rule.png
```
fabianfrz: the image should be IPv4+IPv6 to be future proof

MichaelDeciso: The manual seems to be IPv4-only throughout, currently -- it even disables IPv6 explicitly.

That could be changed, obviously, but it would involve much more than just changing this screenshot, and I would like @AdSchellevis to OK that first.

AdSchellevis:
@MichaelDeciso @fabianfrz MichaelDeciso  IPv6 comment is valid, but this looks like an improvement compared to the current version.  Feel free to test and add IPv6 support in a new pull request.

PR #131, opened 2019-02-07
Fix text wrapping in tables

MichaelDeciso:
This makes the text in tables wrap around if it gets too wide, rather than widening the table and adding a horizontal scrollbar.
Fix is taken from here: https://rackerlabs.github.io/docs-rackspace/tools/rtd-tables.html

voor
na


AdSchellevis:
@MichaelDeciso thanks!

PR #132, opened 2019-02-09
new page: community hardware

flatterlight:
first try ;)
added communityhardware page and linked it in general hardwarepage
with example hardware


AdSchellevis:
the forum is a better place to host this, people can discuss how well options perform and what does and doesn't work, our docs isn't really the place for that. 

Can you move this to the forum please?

PR #139, opened 2019-02-15
Create page about NAT

MichaelDeciso:
This page links to the nptv6 page I wrote earlier (but hasn't been merged yet).

### DIFF
--- a/source/manual/nat.rst
+++ b/source/manual/nat.rst
@@ -15,6 +15,11 @@ addresses are used for the following:
  Redirect      Where the traffic should be redirected.
 ============= ===========================================================================================================
 
+.. warning::
+
+    - Network Address Translation should not be relied upon as a security measure.
+    - Disabling pf will also disable NAT.
+
 --------------------
 Some terms explained
 --------------------
--- a/source/manual/nat.rst
+++ b/source/manual/nat.rst
@@ -39,6 +39,8 @@ succeed since the firewall will not know where to send the traffic. This can be 
 forwarding rules. For example, for a web server behind the firewall to be accessible, ports 80 and 443 need to
 be redirected to it.
 
+Port forwarding is also referred to as “Destination NAT” or “DNAT”.
+
 In OPNsense, port forwarding can be set up by navigating to **Firewall->NAT->Port Forward**. Here, you will see
 an overview of port forwarding rules. New rules can be added by clicking **Add** in the upper right corner.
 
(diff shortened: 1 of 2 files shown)


source/manual/nat.rst
```diff
@@ -0,0 +1,139 @@
...
+ Source
+ Destination / invert   Invert match in “Destination” field.
+ Destination
+ Description            A description to easily find the rule in the overview.
+ NAT reflection         See “Some terms explained”. Leave this on the default unless you have a good reason not to.
+====================== =================================================================================================
+
+--------
+Outbound
+--------
```
fabianfrz: I would add the terms SNAT / Source NAT here as it is commonly called this way (also in netfilter on Linux).

source/manual/nat.rst
```diff
@@ -0,0 +1,139 @@
...
+instead of the internal one (which would the most logical), NAT reflection can rewrite this request so that it uses
+the internal IP, in order to avoid taking a detour and applying rules meant for actual outside traffic.
+
+**Pool options**: When there are multiple IPs to choose from, this option will allow regulating which IP gets used.
+The default, Round Robin, will simply distribute packets to one server after the other. If you only have one external
+IP, this option has no effect.
+
+---------------
+Port forwarding
+---------------
```
fabianfrz: I would add the terms DNAT / Destination NAT here as it is commonly called this way (also in netfilter on Linux).

Should also mention transparent proxies

source/manual/nat.rst
```diff
@@ -0,0 +1,139 @@
+===========================
+Network Address Translation
+===========================
+
+Network Address Translation (abbreviated to NAT) is a way to separate external and internal networks (WANs and LANs),
+and to share an external IP between clients on the interal network. NAT only works on IPv4. For IPv6,
+:doc:`Network Prefix Translation <nptv6>` can be used instead.
```
fabianfrz: NAT64 does exist as well in IPv6

source/manual/nat.rst
```diff
@@ -0,0 +1,139 @@
+===========================
+Network Address Translation
+===========================
+
```
fabianfrz: please add a warning block: "NAT is not a security feature or firewall replacement!".

Maybe also mention that disabling PF will stop NAT as well.

source/manual/nat.rst
MichaelDeciso: I added the terms. What would you like me to say about transparent proxies?

fabianfrz: Please simplify:

All transparent proxies in OPNsense are implemented in a way, that you need to DNAT a connection to a daemon listening on localhost, which tries to get the original desination IP from the `/dev/pf`device. The transparent proxy tries to connect to the original server depending on the proxy configuration.

For example if you want to to use a transparent proxy for HTTP, you need to forward traffic to from TCP/80 via IPv4  to 127.0.0.1:3128 in the default configuration.

MichaelDeciso: Ok, I have a question: why does this fall under "port forwarding"/inbound NAT and not under outbound/SNAT? Is it not possible to act as a transparent proxy for LAN clients?

MichaelDeciso:
This PR should be ready to go.

PR #143, opened 2019-02-21
Correct opnsense-update help command

MichaelDeciso:
`opnsense-update -help`, `opnsense-update --help` or `opnsense-update -h` do not work. As running it already says, the correct way is `man opnsense-update`.


MichaelDeciso:
Note: please squash-merge and make sure the commit message reads `Correct opnsense-update help command`.

PR #147, opened 2019-02-27
Write page about interfaces

MichaelDeciso:


source/manual/interfaces.rst
```diff
@@ -0,0 +1,116 @@
...
+Assignments
+-----------
+
+Most interfaces have to be assigned to a physical port. By default, LAN is assigned to port 0 and WAN is assigned to
+port 1. Assignments can be changed by going to :menuselection:`Interfaces --> Assignments`. This lists existing
+interfaces, with the interface name on the left and the physical port selected in the dropdown. New ones can be
+created here as well, using a non-occupied physical port.
+
+When creating a new interfaces or changing an existing one (which can be done by clicking its name in the overview),
+the following options are available:
```
fabianfrz: You should add a heading as here starts a new page

source/manual/interfaces.rst
```diff
@@ -0,0 +1,116 @@
...
+
+=========================== ===================================================================================================================================================
+ Option                      Explanation
+=========================== ===================================================================================================================================================
+ **General configuration**
+ Enable interface            Disable the interface without needing to remove its assignment altogether.
+ Prevent interface removal   Can be especially useful for important interfaces that rarely ever change, like WAN and LAN.
```
fabianfrz: It is mainly used for virtual Interfaces like tinc or other VPNs

source/manual/interfaces.rst
```diff
@@ -0,0 +1,116 @@
...
+ **General configuration**
+ Description                 A short description for clarifaction, if the name does not suffice.
+ Block private networks      Block traffic claiming to come from private addresses. On WAN interfaces, this kind of traffic should not happen legitimately.
+ Block bogon networks        Block traffic claiming to come from invalid (reserved) addresses (Martian packets). This kind of traffic should not happen legitimately.
```
fabianfrz: -> routing protocols and other multicast

source/manual/interfaces.rst
MichaelDeciso: I don't quite understand. I did start a new section with “Assignments”.

source/manual/interfaces.rst
MichaelDeciso: Could you elaborate?

fabianfrz: for example, bogons block multicast (224.*+), which is used by OSPF, RTMP etc. For that reason the FRR plugin contains a firewall rule to get around that.

source/manual/interfaces.rst
fabianfrz: 1) Assignments
2) Configuration

MichaelDeciso: My approach is consistent with the rest of the documentation. It takes people to the overview, and then explains the options when adding a new entry or modifying an existing one.

source/manual/interfaces.rst
```diff
@@ -17,23 +17,23 @@ created here as well, using a non-occupied physical port.
...
+=========================== ============================================================================================================================================================
  **General configuration**
  Enable interface            Disable the interface without needing to remove its assignment altogether.
  Prevent interface removal   Avoids accidentally removing an interface (has to be turned off and saved before an interface can be removed).
  **General configuration**
  Description                 A short description for clarifaction, if the name does not suffice.
  Block private networks      Block traffic claiming to come from private addresses. On WAN interfaces, this kind of traffic should not happen legitimately.
- Block bogon networks        Block traffic claiming to come from invalid (reserved) addresses (Martian packets). This kind of traffic should not happen legitimately.
+ Block bogon networks        Block traffic claiming to come from invalid or reserved addresses (Martian packets). Note that this also includes multicast traffic over OSPF and RTMP.
```
fabianfrz: both of them are protocols - I would not write 'over' since OSPF and RTMP are encapsulated in IP.

For the FRR plugin, there is a firewall hook to bypass the bogon setting: https://github.com/opnsense/plugins/blob/master/net/frr/src/etc/inc/plugins.inc.d/frr.inc#L43

You can find some infos here: https://en.wikipedia.org/wiki/Multicast_address

MichaelDeciso: I have tweaked the wording. Could you take a look to see if that's sufficient? (If it isn't, do you have a suggestion?)

PR #148, opened 2019-02-27
Set the MTU to 1480

sachaz:


AdSchellevis:
this doesn't really help, it needs context of why you should do that, now the screenshot differs from the settings provided without a good idea of why one should define something else than standard here.

PR #164, opened 2019-04-11
Slightly improved LDAP how-to

koelle25:
I moved the example credentials in the table so they appear as a single cell (as opposed to the current state where it is scattered on two cells).

Additionaly I changed the terms VPN and LDAP to full uppercase consistently throughout the document.


fichtner:
Merged, thank you! :)

PR #167, opened 2019-04-30
wireguard: rename server to local 

mimugmail:
Before going stable with the plugin, update to latest wording


fichtner:
Merged, thanks!

PR #30, opened 2018-07-10
self-signed-chain

qinohe:
self-signed-chain added

PR #188, opened 2019-07-25
Update for 19.7.1

qinohe:
Hey guys, updated this page so it's up to date with 19.7.1
I decided to not remove but to give notes because there are users using older versions.
When it becomes obsolete we can remove the references, at least that seems the best way to me, correct me if I'm wrong ;)


qinohe:
Oh and btw. how do I update my own repo before adding these changes, I see there are conflicts, and they are there because I simply copied the file which is probably not how I supposed to do this ;) Thanks

fichtner:
Hmm, it's hard to tell what is what with a fresh file copied over an existing one. Personally, I'd drop all mentions of before/after 19.7.1 because they are relevant yesterday and then rapidly decline to become relevant over the months and years to come :)

fichtner:
PS: Are you using command line git?

qinohe:
> Hmm, it's hard to tell what is what with a fresh file copied over an existing one. Personally, I'd drop all mentions of before/after 19.7.1 because they are relevant yesterday and then rapidly decline to become relevant over the months and years to come :)

That what you get get if someone not really familiar with git is helping out, sorry ;-)
Right, you may be right yes, I'll go ahead and remove the tweaks needed for before 19.7.1

> PS: Are you using command line git?

Yes, I do use CLI git.
Only the pull is done from the site, 'my repo'

qinohe:
Aha, I think I found a way to solve this:
On page https://github.com/opnsense/docs/pull/30 there's a button: Delete fork.

Should I use this (if I still can..) and than create a new clone?
Or should I continue and create a new pull with the new changes?
Thanks, mark

source/manual/how-tos/self-signed-chain.rst
```diff
@@ -0,0 +1,348 @@
...
+
+Should you even consider using **self-signed certificate chains** in this age of free available
+certificates?
+
+   * Self-signed certificate are just as secure as real world certificates.
+   * They are trustworthy chains, you **know** all parties.
+   * Intranets are often set up with these chains.
+   * Depending on what you do with your network/servers this is a good solution.
+   * Creating your own chain will give some insight in the process.
+   * Only use them if you are sure you can. Read about the concept in common there is a lot of info on the net.
```
fabianfrz: I would like to insert a line
```suggestion
   * Only use them if you are sure you can. Read about the concept in common there is a lot of info on the net.
   * They are required for intercepted connections (see proxy chapter)
```

source/manual/how-tos/self-signed-chain.rst
```diff
@@ -0,0 +1,348 @@
...
+
+What you should not do with a self-signed chain:
+
+   * Use them for a web-facing server.
```
fabianfrz: Insert a line:
```suggestion
   * Use them for a web-facing server.
   * Intercept encrypted traffic on a public wireless network (for example if you provide access in a hotel)
```

source/manual/how-tos/self-signed-chain.rst
```diff
@@ -0,0 +1,348 @@
...
+
+What you should know about self-signed certificates:
+
+   * They are **only** as trustworty as the person, company or organization signing it.
```
fabianfrz: The same is true for certificates accepted by CAB-Forum. There were already some rogue CAs which were in the default store. For example google found some attacks: https://www.zdnet.com/article/google-catches-french-govt-spoofing-its-domain-certificates/

I would have kicked such certificate authorities immediately out of the store but sadly this did not happen.

source/manual/how-tos/self-signed-chain.rst
```diff
@@ -0,0 +1,348 @@
...
+   * Using these certificates **can** be a security risk if you are the one trusting them and not a CA.
+
+A chain will need at least a CA and certificate; an intermediate CA is not needed, but in case of a
+compromise the CA key would be compromised too.
```
fabianfrz: you should elaborate that more. Like why they should be stored on different machines etc.

source/manual/how-tos/self-signed-chain.rst
```diff
@@ -0,0 +1,348 @@
...
+ **Common Name**        internal-ca                         *Default is fine, change to liking*
+====================== =================================== ========================================
+
+.. image:: images/CA.png
+   :width: 100%
+
+.. Tip::
+
+    Always use valid email addresses for your certificates.
+    Bogus addresses can pose a security risk – and not only for certificates.
```
fabianfrz: I don't understand that point and I work with TLS…

The only purpose of the email field is to contact somebody in case of problems and to identify the owner.

source/manual/how-tos/self-signed-chain.rst
```diff
@@ -0,0 +1,348 @@
...
+====================== =================================== ========================================
+
+.. image:: images/webgui-cert.png
+   :width: 100%
+
+.. Tip::
+
+    When creating the server certificate make sure the **CN - common name**
+    is in fact the the **FQDN - Fully Qualified Domain Name**.
+    You can find it on **Linux/Unix** with this command ``hostname -f``
```
fabianfrz: The CN is deprected (or in case of google chrome for example unsupported), the SAN should be used instead and all possible hostnames should be added there.

source/manual/how-tos/self-signed-chain.rst
```diff
@@ -0,0 +1,348 @@
...
+Open your browser and open the OPNsense/webgui page. You should be presented with a certificate that is
+verified by your intermediate CA.
+
+
+---------------------------------------
+A Chain for Your Local Nextcloud Server
+---------------------------------------
+
+
+The local chain for Nextcloud server so we can use OPNsense backup to Nextcloud.
```
fabianfrz: please fix the grammar here. BTW: This is techdoc. Try to get out pronouns like "we" and "you".

source/manual/how-tos/self-signed-chain.rst
```diff
@@ -0,0 +1,348 @@
...
+   cp nextcloud.pem nextcloud.crt
+
+-  * Or use the next quick and dirty method for a single key/certificate file:
+   * Upload the ***.p12**  archive to your Nextcloud server, in a safe way..
+   * Extact the archive into a single **PEM** file and create a certificate.
+
+::
+
+    openssl pkcs12 -in nextcloud-crt.p12 -out nextcloud-crt.pem -nodes
+    cp nextcloud-crt.pem nextcloud-crt.crt
```
fabianfrz: I prefer just pushing the crt and the key file using sftp. But I think this is a question of taste. I don't like to go the long way over the p12.

source/manual/how-tos/self-signed-chain.rst
```diff
@@ -0,0 +1,348 @@
...
+
+-  * **/etc/ssl/localcerts** will be alright for the certificate or choose your own prefered location.
+   * If the key was extracted separatly, **/etc/ssl/private** would be a good choice.
+   * Be sure to set sane permissions on the private directory, ``700`` would do it.
```
fabianfrz: stupid idea. The directory needs to be listable by everyone because many tools need access to some private keys. The keys itself need to be only readable by the right daemon and thy need only 0400.

If you would do that, you could not have a separate key for postfix and nginx for example.

source/manual/how-tos/self-signed-chain.rst
```diff
@@ -0,0 +1,348 @@
...
+   * You could set ``umask`` too (see) ``man umask`` - on your Linux box.
+   * Edit the webserver config to use the certificate and key or single key-cert file.
+   * Sane permissions, ``400`` read only owner is sufficent.
+
+You should now be able to backup to nextcloud and have a verified page.
```
fabianfrz:
```suggestion
You should now be able to backup to Nextcloud and have a verified page.
```

source/manual/how-tos/self-signed-chain.rst
```diff
@@ -0,0 +1,348 @@
...
+.. image:: images/trust.png
+
+Some entries in the form are showed here. Click on the thumbnail for a picture.
+
+When you are done save the form, the CA is now generated.
+
+====================== =================================== ========================================
+ **Descriptive name**   opnsense-ca                         *Choose a name that makes sense to you*
+ **Method**             create an internal ca               *Main purpose of CA*
+ **Common Name**        internal-ca                         *Default is fine, change to liking*
```
fabianfrz: add a long lifetime here for CA certificates. Especially if you build certificate trees.

source/manual/how-tos/self-signed-chain.rst
qinohe: Sadly, yes. Thanks for the link.

Would you like me to add that?
Something like: this is also the case for trusted certificates.

source/manual/how-tos/self-signed-chain.rst
qinohe: Please, add the way you like to add your certs, I will add it. Thanks,

source/manual/how-tos/self-signed-chain.rst
qinohe: I will change that.

source/manual/how-tos/self-signed-chain.rst
qinohe: So, 0777 would do it I think or is that overdone?

source/manual/how-tos/self-signed-chain.rst
qinohe: Well, I had a reason for that, I know 'some' people that would fill in something like user@example.com. Now if I claim example.com the mail would end up there, which could potentially lead to problems...

source/manual/how-tos/self-signed-chain.rst
qinohe: I'm sorry, I would elaborate but I'm not quite sure what I should add here...

source/manual/how-tos/self-signed-chain.rst
qinohe: You mean add a new picture which shows a 10 year lifetime?
Or:
**Lifetime(days)**             3650

qinohe:
@fabianfrz , Thanks for your input and going over the document again. I'm just a hobbyist and IT is not even my profession, so I'm very happy someone is helping me to make this document bulletproof ;-)
I'm not running an email server myself, but if you would like (and have the time) I encourage you to add a piece about an email chain. Of course just suggesting ;-)
Thanks, mark

source/manual/how-tos/self-signed-chain.rst
qinohe: I'm not quite sure how I will change that throughout the document but understand what you mean ;-) Do you also like all pictures changed or just in words?
BTW. I mention that in a Note on line 46-47...

source/manual/how-tos/self-signed-chain.rst
fabianfrz: You don't need my full text but yes, I would mention that.

@fichtner I don't know if we currently have it, but an untrust feature would make sense.

source/manual/how-tos/self-signed-chain.rst
fabianfrz: How about:

The private key of the CA should be stored offline on an USB Stick which you put into a safe so it is not reachable by malicious software or criminals. The intermediate CA, which is intended for a shorter lifetime can be kept on the firewall host.

Since your clients trust the root CA, the root CA can revoke the intermediate in case of compromise.

source/manual/how-tos/self-signed-chain.rst
fabianfrz: Text is enough. Sometimes they are even longer valid like up to 30 years. I think 20 is also common for root CAs.

source/manual/how-tos/self-signed-chain.rst
fabianfrz: Then please add that to the doc.

source/manual/how-tos/self-signed-chain.rst
fabianfrz: I would just write any of your hostnames you added in the SAN field as the server name may not be even matching the DNS names. My VPS is such an example.

source/manual/how-tos/self-signed-chain.rst
fabianfrz: 0755 for directories.

It may start as root and drop privileges so it is not affected but it should always be able to read its own keys.

fabianfrz:
# postfix (main.cf)

```ini
smtpd_tls_cert_file = /path/to/cert.pem
smtpd_tls_key_file = /path/to/key.pem
```

# dovecot (conf.d/tls.conf)

```ini
ssl = yes
ssl_cert = </path/to/cert.pem
ssl_key = </path/to/key.pem

```

source/manual/how-tos/self-signed-chain.rst
fabianfrz: Just download the key and the certificate separately and then do:

```sh
scp -i ~/id_ed25519 /path/to/private/key1.key /etc/ssl/keys/key1.key
scp -i ~/id_ed25519 /path/to/cert1.pem /etc/ssl/localcerts/cert1.pem 
```

With an SSH agent, you don't even need to add the -i private key reference. It is just like copying locally.

qinohe:
@fichtner , are you willing to close the branch at this point so I can delete my fork and start with a new one?
Otherwise with the added commits here which I don't know how to pull myself locally it becomes a mess...
Or you must have a solution of course ;-)
Thanks, mark

qinohe:
> # postfix (main.cf)
> 
> ```ini
> smtpd_tls_cert_file = /path/to/cert.pem
> smtpd_tls_key_file = /path/to/key.pem
> ```
> 
> # dovecot (conf.d/tls.conf)
> 
> ```ini
> ssl = yes
> ssl_cert = </path/to/cert.pem
> ssl_key = </path/to/key.pem
> ```

Thanks @fabianfrz , I have not done anything with it yet, I first took some quick course on git... Hope that worked out well ;)

fichtner:
@qinohe can you resolve the conflicts manually? I only see "master vs. master" which isn't helpful :)

https://github.com/opnsense/docs/pull/188/conflicts

qinohe:
@fichtner I'm away today, so I'll have a look at it tomorrow. I see I need to do a course on solving conflicts too as I looked at that page before but don't know how to do that, yet ;)

qinohe:
Looking back @fichtner , I'm glad this happened what amazingly tool git is, learned a lot today ;)

fichtner:
Is everything addressed now? If not I'd like to merge and see more smaller PRs for the remaining feedback. @fabianfrz @qinohe

fabianfrz:
@fichtner I think not all of my comments have been addressed but that should not be a reason to block it.

fichtner:
Great, merged. Thanks so far ❤️

PR #191, opened 2019-08-06
Update carp.rst

mimugmail:
Make sync with higher skew clearer ( german reference: https://forum.opnsense.org/index.php?topic=13717.msg63207#msg63207 )


AdSchellevis:
@mimugmail thanks!

PR #197, opened 2019-09-06
Add some information to build documentation to README.md

banym:
Added the information that python27 is needed to build the documentation
Added the information how to setup a basic virutalenv to the README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ rights, trademarks etc. and may require the consent of a third party or the
 license of these rights. Deciso B.V. does not represent or make any warranties
 that it owns or licenses any of the mentioned, nor does it grant them.
 
-####Prepare python environment
+#### Prepare python environment
 You need python 2.7 to build the documentation.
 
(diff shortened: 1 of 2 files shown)


AdSchellevis:
@banym it should work on both, maybe better to find out why python3 has issues if it doesn't work, python2 is eol in 2020.

banym:
Hi, you're correct, it builds with python3. But the requirements.txt does not pin the Sphinx version correctly. It looks like the build fails with everything newer than Sphinx 1.8.5.
By default if you run the pip install -r requirements.txt. with python 3.6 or 3.7 it would pull Sphinx 2.2.0 which does not build the documentation. Details on this will follow later. Maybe I will provide a fix for that if I can figure out what function was deprecated or why it fails. 

For now this pull request can be deleted will open a new one with a fix for the requirements.txt for now and review the main problem with Sphinx 2.x

AdSchellevis:
@banym thanks in advance, it would be nice to have a correct set of requirements for python3.

PR #198, opened 2019-09-19
add pfSync hints from backup to master in carp.rst

tk-wfischer:
The state table synchronization should be done from backup to master, too.
Details see https://forum.opnsense.org/index.php?topic=14237.0

The example configuration files would need to be updated, too. As updating those is more work, and is I assume very often necessary when the description in carp.rst is updated, it would be maybe easier to maintain carp.rst when the configuration files would be removed. But I'm not sure about this, so I have only updated the text for the pfSync paragraph.

### DIFF
--- a/source/manual/how-tos/carp.rst
+++ b/source/manual/how-tos/carp.rst
@@ -203,17 +203,22 @@ in our example (on the master server):
 +--------------------+----------------+
 
 ---------------------------------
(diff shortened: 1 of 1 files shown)


AdSchellevis:
It should usually also work with both set to multicast (empty), but when using an address on the master, it should indeed be configured on both. We seem to have missed that in the docs, thanks!

PR #199, opened 2019-09-29
Fixed doc regarding tunnel adress for local instance

marr1977:
In step 1 the documentation says

> What you receive is what WireGuard calls Allowed IP for your local instance, so edit your instance again, remove the value of Allowed IPs you used when setting up and change it to the one you got.

However, there is no "Allowed IPs" field for the local instance. I believe what should be edited is the "Tunnel Address" field of the local instance.


marr1977:
Ok, reverted my change for the first sentence

marr1977:
Was my change ok @mimugmail ?

fichtner:
Merged, thanks!

PR #200, opened 2019-10-03
add Quectel EG25-G modem to list of tested devices

tk-wfischer:
The Quectel EG25-G modem has been successfully tested with OPNsense 19.7:
https://www.thomas-krenn.com/en/wiki/OPNsense_LTE_connection#Configure_Modem

### DIFF
--- a/source/manual/interfaces.rst
+++ b/source/manual/interfaces.rst
@@ -138,6 +138,7 @@ Tested devices by the OPNsense team include:
 
 * **Huawei ME909u-521** (device cuaUx.0)
 * **Huawei E220** (device cuaUx.0)
+* **Quectel EG25-G** (device cuaUx.2)
 * **Sierra Wireless MC7304** (device cuaUx.2) [as of OPNsense 16.7]
 
 .. Note::


AdSchellevis:
Maybe it's better to check https://docs.opnsense.org/manual/how-tos/cellular.html as well for completeness, I think the proper pointers for Huawei and Sierra wireless are in there as well. 

Would you be so kind to check if our documentation is missing anything for a smooth setup?

Thanks in advance.

tk-wfischer:
Thank you for the hint.

I could add entries with the different /dev/cuaU0.* devices in the table here: https://docs.opnsense.org/manual/how-tos/cellular.html#step-2-configure-point-to-point-device

What do you think?
In case that you agree to do it that way, I'd have to alter the rst table so that there are multiple lines for the different possible link interface entries. Should I do this like described here? http://docutils.sourceforge.net/docs/ref/rst/restructuredtext.html#grid-tables

AdSchellevis:
If the link is basically the only parameter changing for the different types, it might be better to add a table to the first section "Supported devices" containing their link settings, in which case the sentence "The screenshots in this guide are for a Huawei ME909u-521 miniPCIe cellular modem." can move to a "Note" widget.

The tables are indeed constructed as mentioned in rst docs.

AdSchellevis:
@tk-wfischer are you still planning to work on this? or should we close the issue for now?

tk-wfischer:
Thank you @AdSchellevis for reminding me, I'll look at it next week and keep you updated.

AdSchellevis:
@tk-wfischer let's close this one for now and reopen when/if you find the time for it.

PR #202, opened 2019-10-23
Give a heads up about whitespace

sevan:
Leading or trailing whitespace is not trimmed from text fields on save in the opensense webui when pasting in the email address.
This can lead to confusion as to why the process fails (should the behaviour in the webui should be changed, instead of documenting a shortcoming instead?)

### DIFF
--- a/source/manual/how-tos/cloud_backup.rst
+++ b/source/manual/how-tos/cloud_backup.rst
@@ -76,7 +76,9 @@ First we need to have a project in the google developer console:
       -  Download the key and save it (for your own use)
       -  Click "Generate new P12 key" and download the key (for your own
          use, you need this one later)
-      -  Copy Email Address, you need it later.
+      -  Copy Email Address, you need it later (be careful to avoid copying
+         leading or trailing whitespae as this will cause spurious failure in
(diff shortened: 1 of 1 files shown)


fichtner:
Hey @sevan,

Hope you are doing well?! :)

It's indeed a good spot to document. One anecdote is that there was a trim() in a password field at one point which is suboptimal by most standards so we try to avoid manipulation.

Although maybe a focus on "valid mail address" instead of "excess whitespace" is a more reasonable approach from a code perspective. What do you think?


Cheers,
Franco

sevan:
Hi Franco,
It's been a while! I'm good, well done for OPNsense, nice work :)
I agree with you about the password field but for the email address, I pointed a colleague at the docs and they progressed through it without a working setup in the end. It was only through searching the forums that I found a suggestion to check for whitespace and indeed removing the excess whitespace resolved it.

dpieski:
I agree with @sevan regarding whitespace on email addresses. I had this issue yesterday and it took me quite a while to figure out there was a space at the end of an email address.

sevan:
@fichtner what do you reckon?

PR #204, opened 2019-10-27
www/nginx: replace image by a better one

### LINKED ISSUE #203: Update NGINX how-to with relevant image
In section 3, the image is just a blank location window. Can we get it updated to show whatever is needed to follow along?

Also, Is that the entirety of everything necessary to setup NGINX? No ports need to be adjusted in NAT, etc?

docs/source/manual/how-tos/nginx.rst

fabianfrz:
* old picture is used somewhere else, so this adds a new one.

closes #203


AdSchellevis:
@fabianfrz thanks!

PR #207, opened 2019-10-29
Update NAT port forward settings list

dpieski:
If TCP/IP version is IPv4+IPv6, the redirect target IP has to include the CIDR when using the combined. This results in 127.0.0.1/32 or ::1/128. One or the other can be put in the Target IP field.

### DIFF
--- a/source/manual/how-tos/proxytransparent.rst
+++ b/source/manual/how-tos/proxytransparent.rst
@@ -58,9 +58,6 @@ left of the **Enable Transparent HTTP proxy** option and click on **add a new fi
 
 The defaults should be alright, just press **Save** and **Apply Changes**.
 
-.. NOTE::
-  If you use **TCP/IP Version**: IPv4+IPv6, the **Redirect target IP** should be either 127.0.0.1/32 or ::1/128. Both are loopback addresses for localhost and either one can be used. You can only enter one in the Target IP field. 
-
 Step 4 - CA for Transparent SSL
 --------------------------------------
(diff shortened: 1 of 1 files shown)


AdSchellevis:
@dpieski the set of defaults is indeed not complete, I don't agree on the note however. The target should either be IPv4 or IPv6 when using an address.

dpieski:
Are you saying that you shouldn't use "**TCP/IP VERSION** IPv4+IPv6"?

AdSchellevis:
yes, it probably won't pass the validations in the gui with an ipv4 or ipv6 target anyway. (if it would, pf would choose the same as the target specified, which is one family)

dpieski:
Ah ok, so should the note advise against selecting "TCP/IP VERSION IPv4+IPv6"?

If you select IPv4+IPv6 and only enter an IPv4 address without the CIDR, it will say you have entered an invalid IPv6 address, similarly if you enter an IPv6 address without the CIDR it will say you have entered an invalid IPv4 address, however, if you add the CIDR to either the IPv4 or IPv6 address it will pass validation and work. I do not know how to tell which family pf chooses though.

AdSchellevis:
just leave the note out and your fine :)

dpieski:
I removed the note. 

Question for you though, for Transparent Proxy, should you have a total of 4 Rules? (1) IPv4, HTTP; (2) IPv4, HTTPS; (3) IPv6, HTTP; and (4) IPv6, HTTPS?

AdSchellevis:
thanks! yes, you probably would.

PR #217, opened 2019-12-03
Updated 2FA intro and HowTo

ei-ke:
PS: Is it normal that github doesn't display the first image in https://github.com/opnsense/docs/blob/master/source/manual/how-tos/two_factor.rst


AdSchellevis:
@ei-ke thanks for your contribution!

PR #218, opened 2019-12-15
corrected download -> upload in 2nd rule of multi interface/guest net example

sjjh:
If I understood the how-to correctly, the second rule of the "Multi Interface shaping for a GuestNet" example had a wrong description for the 'direction' field


AdSchellevis:
@sjjh thanks, you seem to be right. the target description itself seems faulty too (download vs upload)

PR #221, opened 2019-12-20
unclear formulation

sjjh:
Missing description about how to select interface entry


AdSchellevis:
@sjjh removed the line in https://github.com/opnsense/docs/commit/718681315e6733e21cdceca592d906ee7d7ef1b9, it was intended for a tokeniser (which was used in an older version). Thanks

PR #225, opened 2019-12-20
gender of proxy

sjjh:
I'm no English native speaker, but assume that the proxy is neutral and not male. Please double check with a native speaker to ensure that this change really improves the text and doesn't make it worse!


AdSchellevis:
sounds better indeed, thanks!

PR #233, opened 2020-01-17
improve description on how to create license key

### LINKED ISSUE #232: Tutorial about maxmind geoip license key generation is unclear/outdated
With the latest release 19.7.9 the change "For now, this update brings you a GeoIP database configuration page for aliases which is now required due to upstream database policy changes and
a number of prominent third-party software updates we are happy to see included." is announced. In the web front end help text it says, that the OPNsense documentation contains more information. Searching online, I did find https://wiki.opnsense.org/manual/how-tos/maxmind_geo_ip.html
This how to states "Once you have created an account you’ll need to create a license key. Click in the “My License Key” link and generate a key. Save the key ID somewhere safe!!!", doing so I get asked during the key generation at maxmind if geoip update isused (I assume yes) and which version of geoipupdate is used as starting with version 3.1.1 the key is hased. The OPNsense how to doesn't contain any information regarding this choice.
I would be nice to improve the documentation by
* mentioning that geoip update is used (if that is correct)
* stating if the hashed or unhashed key shall be used

sjjh:
Mention to choose wich version of key creation. fixes #232

### DIFF
--- a/source/manual/how-tos/maxmind_geo_ip.rst
+++ b/source/manual/how-tos/maxmind_geo_ip.rst
@@ -14,7 +14,7 @@ Goto https://www.maxmind.com/en/geolite2/signup and create your account. Note th
 Generate License Key
 ######################
 
-Once you have created an account you'll need to create a license key. Click in the "My License Key" link and generate a key. Save the key ID somewhere safe!!!
+Once you have created an account you'll need to create a license key. Click in the "My License Key" link and generate a key. When promted, choose that you use "geoipupdate" in a version "3.1.1 or later" to generate the hashed version of the key. Save the key ID somewhere safe!!!
 
 You do not need to download the config at this point.
 
--- a/source/manual/how-tos/maxmind_geo_ip.rst
+++ b/source/manual/how-tos/maxmind_geo_ip.rst
@@ -14,7 +14,7 @@ Goto https://www.maxmind.com/en/geolite2/signup and create your account. Note th
 Generate License Key
 ######################
 
-Once you have created an account you'll need to create a license key. Click in the "My License Key" link and generate a key. When promted, choose that you use "geoipupdate" in a version "3.1.1 or later" to generate the hashed version of the key. Save the key ID somewhere safe!!!
+Once you have created an account you'll need to create a license key. Click in the "My License Key" link and generate a key. When asked if you use "geoipupdate",  choose "no". Save the key ID somewhere safe!!!
 
 You do not need to download the config at this point.
 


source/manual/how-tos/maxmind_geo_ip.rst
```diff
@@ -14,7 +14,7 @@ Goto https://www.maxmind.com/en/geolite2/signup and create your account. Note th
 Generate License Key
 ######################
 
-Once you have created an account you'll need to create a license key. Click in the "My License Key" link and generate a key. Save the key ID somewhere safe!!!
+Once you have created an account you'll need to create a license key. Click in the "My License Key" link and generate a key. When promted, choose that you use "geoipupdate" in a version "3.1.1 or later" to generate the hashed version of the key. Save the key ID somewhere safe!!!
```
AdSchellevis: small typo `promted` --> `prompted`. If I'm not mistaken, you can also choose to not use geoipupdate. Since @marjohn56 created this doc, let's ask him what he thinks.

marjohn56: But we don't use geoipupdate AFAIK, so does it matter which one you chose?

sjjh: No idea, as I only tried the one version. If both ways okay (did anybody test?), lets just state it like that?

marjohn56: Just checked - either one works..

sjjh: I'd then suggest to change the text like this:
```suggestion
Once you have created an account you'll need to create a license key. Click in the "My License Key" link and generate a key. When asked if you use "geoipupdate",  choose "no". Save the key ID somewhere safe!!!
```

marjohn56: Does that work too? Didn't test that one. :)

marjohn56: I created keys for both versions of geoipupdate, both worked.

sjjh: Damn. Could you? :) (as someone was saying that we don't use geoipupdate... :D)
The confirmation was stating, that the key was saved hashed -- didn't sound too different to choosing the alternative geoipupdate in version 3.1.1 or later -- but you never know...

sjjh:
well, shall we then formulate it explicitly that it doesn't matter what to choose? Or shall we keep it like I suggested, to choose "no"?
Both ways are fine for me -- I just wanted to prevent fear from anybody taking the wrong choice (as it wasn't documented at all, that there is a choice :) )

marjohn56:
Either way..  Your choice.. :)

sjjh:
being lazy, then just commit it as it is. :D

sjjh:
@AdSchellevis feel free to commit. :)

AdSchellevis:
@sjjh ok, thanks!

PR #234, opened 2020-01-21
fix image caption

sjjh:
image caption for "sharing bandwidth evenly" was referencing the example "reserve traffic for VoIP host/SIP trunk" before.

*Note:* the other captions are not too self explaining either -- not sure if that should be changed as well.


AdSchellevis:
@sjjh thanks!

sjjh:
@AdSchellevis you're welcome. Now I only have to get it running. Unfortunately beside following the how to it doesn't work...: https://forum.opnsense.org/index.php?topic=15336.msg69904#msg69904

PR #236, opened 2020-01-29
Update nat.rst

mimugmail:
https://forum.opnsense.org/index.php?topic=15563.msg71522#msg71522

Auto rules created:
```
nat on em1 inet from (em1:network) to 10.10.10.10 -> 10.10.10.1 port 1024:65535
nat on wg0 inet from (wg0:network) to 10.10.10.10 -> 10.13.13.10 port 1024:65535
nat on em0 inet from (em0:network) to 10.10.10.10 -> 172.24.68.133 port 1024:65535
```

So only XX:network is included.


AdSchellevis:
sounds reasonable, thanks!

PR #238, opened 2020-02-24
Update interfaces.rst

mimugmail:


AdSchellevis:
@mimugmail thanks, the page itself might need more work in the future, these additions will certainly help.

PR #240, opened 2020-02-26
fix formatting

ServiusHack:
Move text out of literal block used for the file path.


AdSchellevis:
maybe it's better to write it as an inline code block, like 

.... :code:\`/usr/local/opnsense/mvc/app/config/services.php\` to wire it all together

The line break is a bit confusing here indeed

ServiusHack:
You're right. I wanted to keep consistency with the paths above but they are not inline with the text. Changed it as you suggested.

AdSchellevis:
@ServiusHack thanks, looks good!

PR #250, opened 2020-04-11
OpenConnect VPN firewall interface mention

aschaap:
I noticed https://github.com/opnsense/docs/pull/35 doing something similar to what I intended (improving the documentation for future users like myself, with perhaps too much network-specific details), so I reduced my addition in trying to maintain the clear separation between the OpenConnect plugin and "basic networking". If there is  a page detailing the basics of setting up a VPN connection was ever created, it would be helpful to new users landing on this page. Alternatively, if linking to external guides is not frowned upon, I found the following helpful: https://www.routerperformance.net/using-openconnect-with-newly-released-opnsense-18-1-1/ .


fichtner:
Merged, thanks!

PR #251, opened 2020-04-17
Update ipv6_tunnelbroker.rst

cwhits:
Some of the settings have been moved since the how to was written. Just ran through this all today and verified everything is working  with slaac.


cwhits:
Also, Netflix appears to allow the HE tunnel now. That was only through some cursory testing from my machine though.

AdSchellevis:
@cwhits thanks!

PR #253, opened 2020-04-23
monit: add second example

mimugmail:
Add another example on how to restart IPsec if a tunnel fails.
Had this at customer site where none of the usual IPsec adjustments helped.

### DIFF
--- a/source/manual/monit.rst
+++ b/source/manual/monit.rst
@@ -30,9 +30,9 @@ save it, then apply the changes.
 
 The fields in the dialogs are described in more detail in the “Settings overview” section of this document.
 
--------
-Example
--------
+---------
+Example 1
+---------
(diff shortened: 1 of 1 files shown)


Franck78:
Hi,
this example is strange
You can read it a hundred times, you can't answer basic question

is this thing 'watching' the IPsec tunnel (  )
OR
is this thing 'watching' some remote server (<remote IP> )

if <remote IP> host is down,  the VPN is or is not the problem !

BUT if <remote IP>  <local_ip> are the names used in the StrongSwan examples, it clarifies thing...

https://docs.strongswan.org/docs/5.9/config/quickstart.html

<remote IP> == remote_addrs, Sun ?
 <local_ip> == Moon ?

PR #255, opened 2020-04-28
Update ipsec-rw.rst

mimugmail:
Remove "Road Warrior" as this term seems to be unkown to users no longer than aournd 5 years working in IT :)

### DIFF
--- a/source/manual/how-tos/ipsec-rw.rst
+++ b/source/manual/how-tos/ipsec-rw.rst
@@ -9,8 +9,8 @@ Intro
 -----
 
 Remate adccess to the company's infrastructure is one of most important and critical services exposed
-to the internet. IPsec Mobile Clients offer a solution that is easy to setup and comptabile with most
-current devices.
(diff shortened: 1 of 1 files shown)


source/manual/how-tos/ipsec-rw.rst
```diff
@@ -8,8 +8,9 @@ IPsec: Setup Remote Access
 Intro
 -----
 
-Road Warriors are remote users who need secure access to the company's infrastructure.
-IPsec Mobile Clients offer a solution that is easy to setup and comptabile with most current devices.
+Remate adccess to the company's infrastructure is one of most important and critical services exposed
```
AdSchellevis, line 11: Remote access :) Although it might be good to keep a reference to road warrior as well, a **lot** of people expect the term road warriors.

mimugmail: Fixed

AdSchellevis:
@mimugmail thanks!

PR #258, opened 2020-05-15
Update IPv6_ZenUK.rst

marjohn56:
Addition for directly send solicit option not visible in GUI


fichtner:
We should handle the present in its simplicity in the handbook, not past or future. The best way is to remove mentions of the option that no longer exists.

marjohn56:
Now I've had to change the picture.. Guess I better do Sky as well.

PR #262, opened 2020-05-23
Updated KVM info

binaryanomaly:
Update regarding KVM that more accurately reflects the current status.


AdSchellevis:
@binaryanomaly thanks!

PR #263, opened 2020-06-02
Updated RDNSS RFC from 6106 to 8106

IsaacFL:
RFC 6106 has been obsoleted by 8106


AdSchellevis:
@IsaacFL you're right, thanks for the update!

PR #264, opened 2020-06-02
Deleted inaccurate AdvRouterAddr line in table

IsaacFL:
The AdvRouterAddr is not set for any of these options. I verified the radvd.conf file and opnsense, correctly, does not set this flag.

This flag, AdvRouterAddr, is only set for use with Mobile IPv6.  From the RADVD.CONF(5) man page:
       AdvRouterAddr on|off

	      When set,	indicates that the address of interface	 is  sent  in-
	      stead  of	 network  prefix, as is	required by Mobile IPv6.  When
	      set, minimum limits specified by Mobile IPv6 are used for	 MinR-
	      trAdvInterval and	MaxRtrAdvInterval.

	      Default: off


fichtner:
Indeed, was removed in 20.1.x ... merged, thanks!

PR #271, opened 2020-07-14
Unbound: Added documentation for advanced configurations

xgcssch:
As a followup to opnsense/core#4208 this is the additional documentation for the unbound extension point in OPNSense.
Best regards
Sönke


AdSchellevis:
Thanks! one note on this specific include is that `/var/` is volatile on memory based system, in which case you can use the plugin system to write the file for you (add a template directory [1] and let configd write the file for you, which also happens after reconfiguring).

[1] https://docs.opnsense.org/development/backend/templates.html

AdSchellevis:
@xgcssch Thanks, this will certainly help people trying to extend the configuration manually!

PR #275, opened 2020-07-24
Rewrote Wireguard Road Warrior manpage to be clearer and more concise

IronicBadger:
Original article made several small assumptions about users prior experience with this plugin and the Wireguard technology as a whole. It also glossed over important details that could catch out new users and it's fairly likely that those reading this page, this could be their first encounter with the technology. Therefore being as explicit as possible, but not verbose was my goal.

This rewrite attempts to remove all instances of 'you' and 'just do X' from the text whilst also adding example configurations and highlighting potential pitfalls (which this author experienced just this week). And ultimately should result in fewer support threads on this topic.

Happy to discuss any changes you'd like.

### DIFF
--- a/source/manual/how-tos/wireguard-client.rst
+++ b/source/manual/how-tos/wireguard-client.rst
@@ -6,7 +6,11 @@ WireGuard Road Warrior Setup
 Introduction
 ------------
 
-WireGuard is a simple, fast and modern VPN using modern `cryptography <https://www.WireGuard.com/protocol/>`. It aims to be faster and simpler then IPsec whilst also being considerably more performant OpenVPN. Initially released for the Linux kernel, it is now cross-platform and widely deployable. It is under heavy development and was included in the Linux kernel v5.6 in `March 2020 <https://arstechnica.com/gadgets/2020/03/WireGuard-vpn-makes-it-to-1-0-0-and-into-the-next-linux-kernel/>`. WireGuard is still experimental and should be used with caution.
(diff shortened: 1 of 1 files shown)


mimugmail:
Thanks for taking the time to put this guide to a more technical level :)
Did you test the part of gateway code? The guide was written at 19.7 and with 20.1 gateway code changed so I think this is already outdated.

In the forums I tend to advise this one, but maybe it's working again the old way:
https://forum.opnsense.org/index.php?topic=17383.msg78925#msg78925

IronicBadger:
Hi Michael,

What purpose does the gateway actually serve? I included it because everyone else did, but that doesn't seem like the best reason to include it. Internet access works both with and without a gateway specified (I just tested). So maybe I could remove that section?

I spun up a blank instance as I was writing so that I got everything accurate against `OPNsense 20.1.8_1-amd64`.

mimugmail:
The gateway is needed if you want (need) to do PBR via Firewall rules. 

Many ppl. have let's say a guest network which should go out to WAN and a let's say, Torrent net, which should go out to Azire or whatever.

IronicBadger:
@mimugmail Is that within scope of this configuration? The use case here is presumed to be a client connected to OPN from somewhere on WAN and routing all traffic via that tunnel.

Perhaps a separate article would be better for those more complex scenarios?

source/manual/how-tos/wireguard-client.rst
```diff
@@ -6,86 +6,177 @@ WireGuard Road Warrior Setup
...
+If more granular rules are required note there is a new interface **wg0** where these may be configured.
 
-Now we can **Enable** the VPN in tab **General** and continue with the setup.
+The final piece is to allow traffic from the Wireguard network. Do this via :menuselection:`Firewall --> Rules --> WireGuard` and click **+Add** with the following information (if an item is not specified, leave it set to the default value):
 
-If you want to add more users just add them in **Endpoints** and link them via **Peers**.
+=========================== ================ =====================================================================
+ **Interface**               WireGuard        *The interface this rule applies to*
+ **Source**                  WireGuard net    *Source subnet*
+ **Destination**             any              *Traffic destination*
```
mimugmail, line 97: Destination should be LAN net if you use this wording in description.

AdSchellevis:
@IronicBadger nice work, definitely an improvement, thanks!

PR #276, opened 2020-07-26
Update sslvpn_client.rst

mimugmail:
While helping here https://forum.opnsense.org/index.php?topic=14336.15 I found that the guide refers to User Auth only where no client export is available and also since creating of client certificates is explained here, I'm sure we need to add SSL/TLS here

### DIFF
--- a/source/manual/how-tos/sslvpn_client.rst
+++ b/source/manual/how-tos/sslvpn_client.rst
@@ -238,7 +238,7 @@ For our example will use the following settings:
 
 ===================================== ===============================================
  **Description**                       *My SSL VPN Server*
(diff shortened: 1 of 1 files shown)


AdSchellevis:
@mimugmail although it can be confusing, I think the intend of the document was to start with "user auth" and pin to a client certificate in the last step (Multi Factor Authentication).

mimugmail:
Indeed, but it's more confusing when you follow the guide step by step and wont see any certificates to export which is documented also with screenshots. I dont think it hurts or add more support tickets

AdSchellevis:
@mimugmail let's try it out then, we can always revert if this doesn't help.

PR #279, opened 2020-08-06
Update ipsec-s2s-route.rst

mimugmail:
Add notes for NAT in route-based IPsec

https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=248474

### DIFF
--- a/source/manual/how-tos/ipsec-s2s-route.rst
+++ b/source/manual/how-tos/ipsec-s2s-route.rst
@@ -503,3 +503,18 @@ Gateway Site-B
 
 
 Now you are all set!
+
+
+----------------
+Appendix A - NAT
+----------------
+
+Route-based tunnels do not use SPD settings like policy-based tunnels.
+Just set the usual NAT rules on your assigned IPsec interface and the following tunables:
+
+=================================== ===
+ **net.inet.ipsec.filtertunnel**     1
+ **net.inet6.ipsec6.filtertunnel**   1
+ **net.enc.out.ipsec_filter_mask**   0
+ **net.enc.in.ipsec_filter_mask**    0
+=================================== ===


source/manual/how-tos/ipsec-s2s-route.rst
```diff
@@ -503,3 +503,18 @@ Gateway Site-B
...
+----------------
+
+Route-based tunnels do not use SPD settings like policy-based tunnels.
+Just set the usual NAT rules on your assigned IPsec interface and the following tunables:
+
+=================================== ===
+ **net.inet.ipsec.filtertunnel**     1
+ **net.inet6.ipsec6.filtertunnel**   1
+ **net.enc.out.ipsec_filter_mask**   0
+ **net.enc.in.ipsec_filter_mask**    0
```
fichtner, line 519: Woah, let's slow down for a second. These two make IP traffic work with pf (IPsec rules tab). If we set this to zero we can't filter and users will probably complain.

Is "filtertunnel" a replacement or an addition? Would leaving the masks untouched work or are they part of the issue?

mimugmail: In my first test I just tested with filter_mask and without touching filtertunnel. I saw the reply incoming on the entry Firewall (Site A) but not going out to LAN, only after also setting filtertunnel packets walked through

mimugmail: Previous defaults were:

```
root@PB-FW1-KARL:~ # sysctl net.enc.out.ipsec_filter_mask=0
net.enc.out.ipsec_filter_mask: 1 -> 0
root@PB-FW1-KARL:~ # sysctl net.enc.in.ipsec_filter_mask=0
net.enc.in.ipsec_filter_mask: 2 -> 0

```

mimugmail: @fichtner I set the tunables in my lab and added the filter rules on the assigned interface (not that route-based creates a new interface tab) and when I block ICMP .. it's blocked. So user based filtering works

fichtner: Ok, let me test that tomorrow on my end. From the looks of it it means the old hooks are no longer needed and replaced by the new and simpler one?! 😊

fichtner: PS: it would mean we switch the default in 20.7.x and don’t document if @adschellevis approves too.

mimugmail: Nah .. we only set it in the docs for now, luckily route-based IPsec is still not widely used. Maybe for a major like 21.1 :)
If you need a second endpoint to test, you know where to find me

AdSchellevis: The side affects on other setups can be annoying, which kind of pleads for documentation first. I do however think that it's a good idea to add some lines on what we're changing here. These settings have a tendency to be a bit obscure, in which case it would help to explain why we change things.

jimmielightner: I'm curious if this is the cause of the recent issues I've had with Azure route-based IPSec VPN. Was this ever followed up on?

mimugmail: It still doesnt work, bug at FreeBSD still open. I'm watching it :)

PR #281, opened 2020-09-04
Update tor.rst

ccesare:
This small edit clarifies that a Tor exit node is a certain kind of relay. The implication in the current documentation is that all relays are exit nodes. (See: https://www.grammarly.com/blog/which-vs-that/.)


fichtner:
Merged, thanks

PR #283, opened 2020-09-05
update aliases.rst for new feature: Address exclusions

kulikov-a:
update aliases.rst for new feature: Address exclusions


AdSchellevis:
@kulikov-a looks good, thanks a lot!

PR #286, opened 2020-09-30
Update installazure.rst

mimugmail:
Point links to english pages instead of nl-nl


AdSchellevis:
@mimugmail nice catch, thanks

PR #289, opened 2020-10-16
Update proxyicapantivirus.rst

dpieski:
correct typo. The second test should be for the HTTPS


AdSchellevis:
@dpieski nice catch, thanks!

PR #295, opened 2020-12-01
Add section about OpenVPN assigned interfaces

marcquark:
closes https://github.com/opnsense/core/issues/4485

Add instructions how to work around OpenVPN issues with reply-to in firewall rules


AdSchellevis:
@marcquark looks good, thanks!

PR #296, opened 2020-12-26
Updated how-to for Google Drive backups

Greelan:
As suggested in [this thread](https://forum.opnsense.org/index.php?topic=11900.0), the PR is for an updated how-to for setting up Google Drive backups.

Note that none of the Google Project or service account details in the screenshots are sensitive, as they were created only for the purposes of the how-to.

I did my best with the reStructuredText formatting, but unfortunately have not had much experience with it. Hopefully it is usable. :) Some of the screenshots might need to be scaled down (I didn't have much success with that).


AdSchellevis:
@Greelan thanks, I'll take a look. Usually I try to limit the number of screenshots, since visible appearance  seem to change a lot over time, but since the whole procedure seems to be different now I don't mind too much.

Greelan:
No problem @AdSchellevis. I understand your comment about the screenshots - I hesitated initially to include them, at least to the full degree, as it seemed a bit overkill, and somewhat obvious in places. But some of the comments from users on the forum thread suggested that they were struggling with just written instructions without visuals, so... 

Sorry also for the bevy of commits in that PR, I'm not particularly expert with GitHub lol

AdSchellevis:
@Greelan no problem at all, your efforts are much appreciated.

PR #297, opened 2021-01-02
Editing and adding links to IPS page

bderrly:
I found a few of the paragraphs confusing when I read through this so this is my attempt to make them more readable. Of course, I have a hard time stopping once I start so the size of the edit grew quickly.

I think there is still some room for improvement in this document. In particular, the title is about IPS but the default when enabling this service is IDS and yet that is barely mentioned. I think more attention needs to go into defining the systems, when to use which, and how to decide which to use. I have next to zero experience with either of these so I can't write anything authoritative.

### DIFF
--- a/source/manual/ips.rst
+++ b/source/manual/ips.rst
@@ -1,57 +1,86 @@
-==================================
-Inline Intrusion Prevention System
-==================================
+===========================
+Intrusion Prevention System
+===========================
 
-The inline IPS system of OPNsense is based on Suricata and utilizes Netmap to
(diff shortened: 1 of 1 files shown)


AdSchellevis:
I'm not too fond of the new line-wraps, they complicate review on our end, but the general readability seemed to have been improved. thanks!

bderrly:
Yeah, sorry about the line wraps. I should have kept those as two separate reviews but by the time I realized that I had already gone too far and couldn't undo it easily. Thanks for reviewing and approving.

PR #298, opened 2021-01-06
added notes regarding the alias flush function; Fixes #4196

zerwes:
Fixes issue #4196

### DIFF
--- a/source/development/api/core/firewall.rst
+++ b/source/development/api/core/firewall.rst
@@ -32,10 +32,3 @@ Firewall
     "``POST``","firewall","alias_util","flush","$alias"
     "``GET``","firewall","alias_util","list","$alias"
     "``GET``","firewall","alias_util","update_bogons",""
-
-.. Warning::
-
-    Please note thet the **flush** action is not persistent!
-    See: `issue 4196 <https://github.com/opnsense/core/issues/4196>`_
-
-    For now the only way to flush all elements from a alias is to collect the elments using `alias_list`and delete each element in a loop.
--- a/source/development/api/core/firewall.rst
+++ b/source/development/api/core/firewall.rst
@@ -32,3 +32,10 @@ Firewall
     "``POST``","firewall","alias_util","flush","$alias"
     "``GET``","firewall","alias_util","list","$alias"
     "``GET``","firewall","alias_util","update_bogons",""
+
+.. Warning::
+
+    Please note thet the **flush** action is not persistent!
+    See: `issue 4196 <https://github.com/opnsense/core/issues/4196>`_
+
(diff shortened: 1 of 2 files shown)


source/development/api/core/firewall.rst
```diff
@@ -32,3 +32,10 @@ Firewall
     "``POST``","firewall","alias_util","flush","$alias"
     "``GET``","firewall","alias_util","list","$alias"
     "``GET``","firewall","alias_util","update_bogons",""
+
+.. Warning::
```
AdSchellevis, line 36: the api docs are autogenerated, better leave custom notes out of them.

source/manual/aliases.rst
```diff
@@ -72,6 +72,11 @@ As you can see there are multiple IP addresses for this domain.
 Hosts type Aliases can contain exclusion hosts.
 Exclusion addresses starts with "!" sign (eg !192.168.0.1) and can be used to exclude hosts from Network Group Aliases.
 
+.. Warning::
+
+    Please note thet the **Flush** action is not persistent!
+    See: `issue 4196 <https://github.com/opnsense/core/issues/4196>`_ and :doc:`API </development/api/core/firewall>`
```
AdSchellevis, line 78: better not refer to an issue but  explain what the behaviour is, in this case "flush" means flush the current contents of the alias, which will be repopulated when it's not an external type. (so flush in most cases isn't very useful)

source/development/api/core/firewall.rst
zerwes: But this is where I stumbled over the problem. So this is the place I would have been expecting the note to be present and save me some work and self-doubts ...
Is there a better way to add notes to this page?
And in general: is it useful to keep auto-generated files in git?
So, from your point of view: should I revert this part?

source/manual/aliases.rst
zerwes: see 396bde8fd670f83926e9bbb8741d7d35be4d8157

source/development/api/core/firewall.rst
AdSchellevis: you can overwrite the standard template, which is what we do for the firewall plugin for example (https://github.com/opnsense/docs/blob/master/source/development/api/plugins/firewall.rst.in), but to avoid random notes for every challenge a user may step into, I would only document general behaviour on the gui part in this case (its not specific to the api). so, yes, I would revert it and describe what flush does in the aliases section (next note).

zerwes: changes in `source/development/api/core/firewall.rst` reverted

zerwes:
Should I close this PR and reopen a new one with the commits squashed?

AdSchellevis:
nah, we can squash in GitHub. thanks!

PR #302, opened 2021-01-21
Fix Tayga NAT64 how-to

maurice-w:
As reported in opnsense/plugins#2094, the section about Tayga's IPv6 address was wrong.


AdSchellevis:
@maurice-w thanks!

PR #304, opened 2021-01-30
Update wireguard-client.rst to clarify use of WireGuard net

Greelan:
This proposed tweak to the how-to is intended to address the issue discussed in this forum thread: https://forum.opnsense.org/index.php?topic=21170.0

### DIFF
--- a/source/manual/how-tos/wireguard-client.rst
+++ b/source/manual/how-tos/wireguard-client.rst
@@ -89,18 +89,20 @@ For external clients to connect to the WireGuard server firewall rules must be c
 
 If more granular rules are required note there is a new interface **wg0** where these may be configured.
 
-The final piece is to allow traffic from the Wireguard network. Do this via :menuselection:`Firewall --> Rules --> WireGuard` and click **+Add** with the following information (if an item is not specified, leave it set to the default value):
+The final piece is to allow traffic from the WireGuard network. First define an alias (e.g. **VPN_clients**) and include in it the IP addresses (e.g. 10.10.10.2 and 10.10.10.3) or subnet (e.g. 10.10.10.0/24) of the WireGuard clients from which traffic is to be allowed. Do this via :menuselection:`Firewall --> Aliases` (click **+** in the bottom right).
+
+Then create a firewall rule via :menuselection:`Firewall --> Rules --> WireGuard` (click **+Add** in the top right), with the following information (if an item is not specified, leave it set to the default value):
 
 =========================== ================ =====================================================================
  **Interface**               WireGuard        *The interface this rule applies to*
- **Source**                  WireGuard net    *Source subnet - but see the warning below*
+ **Source**                  VPN_clients      *Source subnet - use the alias defined as above*
  **Destination**             any              *Traffic destination*
  **Description**             WG WAN to LAN    *Optional - provide a description*
 =========================== ================ =====================================================================
 
-.. Warning::
+.. Tip::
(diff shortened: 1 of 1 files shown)


fichtner:
@mimugmail ping :)

source/manual/how-tos/wireguard-client.rst
```diff
@@ -93,11 +93,15 @@ The final piece is to allow traffic from the Wireguard network. Do this via :men
...
  **Interface**               WireGuard        *The interface this rule applies to*
- **Source**                  WireGuard net    *Source subnet*
+ **Source**                  WireGuard net    *Source subnet - but see the warning below*
  **Destination**             any              *Traffic destination*
  **Description**             WG WAN to LAN    *Optional - provide a description*
 =========================== ================ =====================================================================
 
+.. Warning::
+
+    Using **WireGuard net** as the Source will not work as expected if multiple local WireGuard peers or multiple WireGuard endpoints are configured. In those cases, define an Alias that includes the IP addresses (for example, 10.10.10.2 and 10.10.10.3) or subnet (for example, 10.10.10.0/24) of the VPN clients that are intended to have LAN access. Then include that Alias as the Source in the firewall rule.
```
mimugmail, line 103:
```suggestion
    Using **WireGuard net** as source will not work as expected if multiple local WireGuard instances or WireGuard endpoints are configured. In those cases, define an alias including the IP addresses (e.g. 10.10.10.2 and 10.10.10.3) or subnet (e.g. 10.10.10.0/24) of the VPN clients that are intended to have access.
```

mimugmail: I'm not a native speaker but would write it like this.

Greelan:
These largely look like stylistic changes, which I am not going to get upset about (even though I am a native speaker 😀). Deleting the last sentence I included does however mean the instructions only explicitly tell the user to define an Alias, but not to use it in the rule. Implicit (and obvious), sure, but some people seem to need things spelled out.

source/manual/how-tos/wireguard-client.rst
Greelan: These largely look like stylistic changes, which I am not going to get upset about (even though I am a native speaker 😀). Deleting the last sentence I included does however mean the instructions only explicitly tell the user to define an Alias, but not to use it in the rule. Implicit (and obvious), sure, but some people seem to need things spelled out.

sovanyio:
Primarily here to say THANK YOU, I have spent more time than I care to admit debugging why I couldn't get wireguard to work. I now have a working setup :)

Second, I think that the note under 2c for `Source address` might still be a bit misleading to some folks.

It might be more clear to have the warning as its own section prior to sections b and c. Perhaps something like `Creating Wireguard Network Alias`. I think it will be far more common that someone sets up a multi-endpoint setup and that the single endpoint/single-server folks should be the exception in the instructions:

`Info: If using a single endpoint with a single server, it is possible to use "Wireguard Net" rather than create an alias.`

Greelan:
Fair comment. I will have a look at some revised wording along those lines.

Greelan:
@mimugmail Do you have any comments on the revised text? Thanks

Greelan:
> Looks good, thank you :)

Thanks. Look forward to it finally being merged - nudge, @fichtner 😛

fichtner:
Merged, thanks! :)  *goes back to sleeping on the desk*

PR #309, opened 2021-02-07
clarify number of Public Services needed

sjjh:
The how-to stated, that only one Public Service is needed, if multiple domains share the same ip. This is true, if we talk about the same port. If we have different ports, afaik we need one Public Service for each port. Elaborated a little on the description, to make it clearer.


AdSchellevis:
@sjjh thanks!

PR #312, opened 2021-02-14
vim-tiny was renamed vim-console

gwjo:
Replace reference to the `vim-tiny` package that was renamed to `vim-console` about 3 years ago.


fichtner:
It’s still a silly name 3 years later. While touching that line we should add emacs-nox as well? That page is scheduled for a rewrite in any case.

PR #314, opened 2021-02-23
Correcting mismatched interfaces in upload rule

SalazarBruno:
The names of the interfaces were in wrong order on the upload rule


AdSchellevis:
@SalazarBruno thanks!

PR #315, opened 2021-02-24
git-backup: hint to cronjobs for shorter backup intervals

marcquark:
From this topic https://forum.opnsense.org/index.php?topic=21639.0

There is no officially documented way of decreasing the default push period for git backups (which is apparently 24 hours). It's super easy to do by setting up a cronjob. So adding a note to the docs doesn't hurt :)


AdSchellevis:
@marcquark you're right, thanks!

PR #318, opened 2021-03-04
API: Add a note for postman users

swhite2:
It may not be directly obvious to NOT use the API key authorization method in Postman. Instead, basic auth is used.


AdSchellevis:
@swhite2 thanks!

PR #322, opened 2021-03-27
WG selective routing how-to: DNS related updates

Greelan:
Address DNS resolution on OPNsense and DNS leaks


mimugmail (approved):
Thx & sorry for the delay :)

PR #330, opened 2021-04-29
Add wiki info for Alias  Dynamic IPv6 Host

marjohn56:
Info on how and why to use Dynamic IPv6 Host Alias


source/manual/aliases.rst
```diff
@@ -13,33 +13,31 @@ Alias Types
...
-| Hosts            | Single hosts by IP or Fully Qualified Domain Name  or|
-|                  | host exclusions (starts with "!" sign)               |
-+------------------+------------------------------------------------------+
-| Networks         | Entire network p.e. 192.168.1.1/24 or network        |
-|                  | exclusion eg !192.168.1.0/24                         |
-+------------------+------------------------------------------------------+
-| Ports            | Port numbers or a port range like 20:30              |
-+------------------+------------------------------------------------------+
-| MAC addresses    | MAC address or partial mac addresses like            |
-|                  | :code:`f4:90:ea`                                     |
```
AdSchellevis, line 28: @marjohn56 oO, where is our MAC address type? ;)

source/manual/aliases.rst
```diff
@@ -190,6 +188,37 @@ Combine different network type aliases into one, this type of alias accepts othe
...
+for the firewall entry that automatically tracks the prefix and changes the rule. Create a new IPv6 Dynamic Host alias and enter
+only the suffix of the address, in the case of a /64 using the above example of the server address you would enter the lower 64
+bits of the address :2800:1000:1000::1/64, note the ':' at the start of the address, you MUST start the address with a ':'.
+
+When the prefix changes, the alias address will then be updated in the firewall rules, let's say your prefix changes to 
+2000:1111:2222:3333:3200::/56 the rule updates and the entry for your server in the firewall would update automatically to be
+2000:1111:2222:3333:3200:1000:1000::1
+
+The IPv6 Dynamic Host alias entry allows you to enter any prefix length you like, but in most cases a /64 will be used. If
+you do not enter a prefix length it will be assumed to be a /64. You may enter multiple addresses, for example if you have
```
AdSchellevis, line 215: @marjohn56 If I'm not mistaken, we enforce mask with our validations.

marjohn56:
Now that's because I did not pull upstream before starting the commit, which proves two things, one my auto script did not work, and two I don't read the wiki very often. I'll correct and update in the morning.

marjohn56:
OK, try that...

source/manual/aliases.rst
marjohn56: I didn't notice that had been removed. My version auto set to /64 if no prefix was added. Doesn't matter, I've removed said text from the file.

maurice-w:
It seems all the example prefixes / addresses are 16 bits too long. :-) Also, may I suggest sticking to 2001:db8::/32 for documentation?

marjohn56:
Hmm.. now I cant count anymore... I'll remove the offending bytes tomorrow. Why that specific address?

maurice-w:
2001:db8::/32 is the "official" prefix reserved for documentation: https://tools.ietf.org/html/rfc3849

marjohn56:
Well, learn something new every day. :)

marjohn56:
OK, check that it meets all requirements pour favour

source/manual/aliases.rst
```diff
@@ -190,6 +193,37 @@ Combine different network type aliases into one, this type of alias accepts othe
...
+For example, you obtain a prefix 2001:db8:2222:2800::/56.  You have a /56 prefix and if the tracking id was set to 0 for your 
+LAN, you would have an address range on your LAN of 2001:db8:2222:2800:: to 2001:db8:2222:2800:FFFF:FFFF:FFFF:FFFF.
+
+You want to run a server on your LAN that is accessable from the WAN so you give it a static address of 
+2001:db8:2222:2800:1000:1000::1 and create a rule allowing traffic to access the server.
+
+When your prefix changes, that static address is no longer valid, so you must use the IPv6 Dynamic Host to create an alias address
+for the firewall entry that automatically tracks the prefix and changes the rule. Create a new IPv6 Dynamic Host alias and enter
+only the suffix of the address, in the case of a /64 using the above example of the server address you would enter the lower 64
+bits of the address ::2800:1000:1000:1/64, note the '::' at the start of the address, you MUST start the address with a '::'.
```
maurice-w, line 213: The lower 64 bits are ::1000:1000:0:1

maurice-w: Same in the screenshot.

AdSchellevis: for a specific address it should probably also use /128

marjohn56:
I've changed it to a draft, so edit as you see fit. You need to add the prefix length, but from my testing only the /64 is used.

AdSchellevis:
@marjohn56 @maurice-w the prefix provided specifies the prefix used for the resulting address, so `::1000/64` would wipe 
the bits specified by the interface address and appends  `::1000/64` to it, which doesn't make sense if you're aiming for a specific address. Just test the output and see what it does in pfTables.

marjohn56:
I've modified the interfaces.py to work in a different way. The user enters an address such as '::1234:5678:9abc:def1:2345/48', and appended that to a /48 prefix from the interface. Now, that produces the correct address, e.g 2001:0000:0000:1234:5678:9abc:def1:2345/48
 Should the aliastable entry stay as a /48 or should it be set to a /64 ?
I know this is the wrong place to be discussing this, but I can open a new issue and/or commit when this question has been answered.

maurice-w:
It seems host aliases and network aliases are getting mixed up here. For static addresses, "Host(s)" and "Network(s)" are separate alias types. Is this new alias type supposed to be both?

An IPv6 _host_ alias is always /128.

An IPv6 _network_ alias can have any length. You need to be able to specify the prefix length of the resulting alias as well as the number of bits of the specified interface's address you want to merge with the suffix.

@marjohn56, an alias like `2001:0000:0000:1334:5678:9abc:def1:2345/48` makes no sense to me. If it's a /48 network, you would just use `2001:0000:0000::/48`. If it's a host, you would make it a /128.

marjohn56:
No, that's what I mean, I think the Alias 'IPv6 Dynamic Host' should be a 128; but wasn't it you who wanted to be able to specify the mask when entering an address? If so then the mod works, all I need to to is pop a 128 on the end instead of the mask entered and you will be able to enter a /48 address.

AdSchellevis:
We probably need to change the name, I don't mind allowing a partial address without a mask (omit /128) as well, but the type itself is indeed a network and not a host. how about "IPv6 dynamic net"?

maurice-w:
What I don't understand: We have separate alias types for static hosts and static networks. For dynamic IPv6 we now only have a network alias and if you want to specify a host you just make it a /128 "network"? It's possible, sure, but I think it makes it very confusing (as this discussion proves).

I suggest having two separate alias types:
"IPv6 Dynamic Host(s)"
"IPv6 Dynamic Network(s)"

If we can only have one:
"IPv6 Dynamic Host(s) / Network(s)"

marjohn56:
Going back to your earlier comment - 

> Does the prefix length specify how many bits of the selected interface's address are merged with the suffix? So if the selected interface has the address 2001:db8:2222:28de:1234:56ff:fe78:9abc and you enter the suffix ::ab:0:0:0:1/56, the resulting alias address would be 2001:db8:2222:28ab::1?

With the change I've made, then that's exactly what happens.

What would you prefer to enter? A host normally lives on a /64 network so a /128 suffix can be added and you only enter the 64bit suffix. No mask would assume a /128 with  a /64 prefix?

AdSchellevis:
the calculation itself is more or less https://github.com/opnsense/core/issues/4923#issuecomment-820573860, if it's more clear for the enduser we might consider splitting types indeed, but then again an address in a current network type entry is also a host....

fichtner:
Huh, I thought the network size indicated how much prefix width we merge. Behind /64 it doesn’t make much sense, /128 would mean we only want the prefix, but the prefix is limited to /64 anyway, otherwise we end up with bits of the source interface address.

The concept is pretty solid. It also allows embedding bits into the /64 if we delegate a /60 and set the 4 bits manually via the suffix alias. Later, as Maurice suggested we could inject the correct prefix id via alias setting too.

marjohn56:
#4981 in core allows a variable length. As  I said, what suffix you attach is erm... variable 😉

AdSchellevis:
I wasn't planning merging additional code as long as we're still seem to discuss what the function should do, the previous discussions assumed we chopped of the provided netmask on the interface and prepended the provided "partial". 
As far as I know that's exactly what it's doing now. 

We can discuss if networks should be supported or not and in which manner (both are likely already supported, validations might need some changes).

marjohn56:
OK, here's the results of a test with the code as it stands, without any changes:

UI Entry:  ::1234:5678:9abc:def1:2345/48

Result:

AliasTable entry: 2xxx:8xxx:6xxx:f900:5678:9abc:def1:2345/48

So it's only taken the /64 suffix and not the full /48

AdSchellevis:
I think we're a bit lost in translation, can someone prepare some full examples of what the address tracking should do? e.g. 

interface  `2001:2c26:f4e4:0:21c:ffff:fefe:4a31/64`
input         `::1004/128`
output `2001:2c26:f4e4::1004`

Explicit examples help debugging and detailing what we all think the new functionality should do and figure out if we all agree on the subject, if we're not able to make things explicit we might need to revert the functionality in core.

marjohn56:
OK, here's how I see what should happen, this is for a host alias only.  I don't have need for or use a network alias so better that someone who does explains their needs.

A user has a dynamic  IPv6 PD, which may change either on reboot or the ISP changes it every 'n' hours.
His LAN prefix therefore changes. Any rule he has created for a host on his LAN will therefore become invalid when the prefix has changed. The object is  therefore to change the prefix of the address to match that of the new prefix obtained. So take this as an example, I have a web server on my LAN, I want to allow traffic in through the fw so I create a rule to allow the connection. The current prefix is a /56:
2001:db8:aaaa:bb/56

So the address of the server, SLA IS is 0 on this interface I enter into the fw rule is:
2001:db8:aaaa:bb00:0000:0000:0000:0010/128

Now the prefix gets changed
2001:db8:aaaa:dd/56

I need to change the prefix on the address of the server to match this so it becomes:
2001:db8:aaaa:dd00:0000:0000:0000:0010/128

There are two ways to look at this, as a host the host mask should be /128, but I cannot currently enter that in the alias, I need to be able to change the full prefix length, so if I do enter /128, I must look at the PD mask obtained and use that as size to split the prefix suffix and automatically make it a /128 in the aliastable ,
**OR**  
I enter the mask size I want, i.e. /56 and assume that because it's a host entry then it will be given a /128 in the aliastable.

That's my take on it.

maurice-w:
@AdSchellevis

Interface address for all examples: `2001:db8:1234:5678:4343:65ff:fe56:7878`

**"IPv6 Dynamic Host(s)":**

Example a):
Input 'Content': `::a`
Input 'Interface bits [1]': `64`
Output (Alias): `2001:db8:1234:5678::a`

Example b):
Input 'Content': `::f:0:0:0:a`
Input 'Interface bits': `60`
Output (Alias): `2001:db8:1234:567f::a`

Example c):
Input 'Content': `::f:0:0:0:a`
Input 'Interface bits': `56`
Output (Alias): `2001:db8:1234:560f::a`

**"IPv6 Dynamic Network(s)":**

Example a):
Input 'Content': `::b/64`
Input 'Interface bits': `56`
Output (Alias): `2001:db8:1234:560b::/64`

Example b):
Input 'Content': `::c0/60`
Input 'Interface bits': `56`
Output (Alias): `2001:db8:1234:56c0::/60`

Example c):
Input 'Content': `::ef00/56`
Input 'Interface bits': `48`
Output (Alias): `2001:db8:1234:ef00::/56`

[1] Probably a dropdown, and need to find a better name for "how many bits of the interface's address to use".

AdSchellevis:
@maurice-w currently we wipe the number of interface specified bits and append the content + netmask, your example suggests there's another "interface bits" then the one offered by the system. I'm a bit puzzled now why that is. Adding more toggles has the downside of making it even easier to get lost in translation.  With the three of us it already seems to be quite difficult to end up with a proper specification for the new feature....

AdSchellevis:
yes, we should either fix this properly and sustainable or leave it out of the product, it should be perfectly clear what the functionality offers and how one should use it. If we only offer a concept that works for a host and haven't thought through how one should use this for a network, one day it's raining support cases which are hard or impossible to solve. 
Not saying we should include both types in the same version, at least there should be a clear vision. Given all the noise we currently seem to be able to generate with only a few lines of code, it's a bit scary what could be next. 

It could very well be I miss some piece of information (don't have a track setup here), in which case we might indeed need additional input, although I still have a feeling that we risk looking at the wrong end and already solved the puzzle but ended up misplacing the pieces.

@maurice-w looking at network example b (`2001:db8:1234:56c0::/60`) which address is (or addresses are) configured on the interface this one is tracking (wan)? 

The alias type itself doesn't state it's only suitable for a tracking interface, although it might be the only use-case we currently know of, it might be hard to predict it's the only one. In my honest opinion it's really worth make sure logic remains consistent and easy to extend.

maurice-w:
Initially only having dynamic host aliases with 64 fixed interface bits is perfectly fine. This will already be a big improvement and cover many use cases. A separate dynamic network alias type could possibly be implemented later. I'm in no position to expect that this gets done anytime soon.
But as Ad noted, it is essential that we define and document very precisely and without ambiguity what the feature is supposed to do and how to use it.

> @maurice-w looking at network example b (`2001:db8:1234:56c0::/60`) which address is (or addresses are) configured on the interface this one is tracking (wan)?

I really tried, but I don't understand the question. We must have some major miscommunication here. :-/ The WAN interface address is unrelated to the delegated prefix.

AdSchellevis:
> I really tried, but I don't understand the question. We must have some major miscommunication here. :-/ The WAN interface address is unrelated to the delegated prefix.

Likely, as I've mentioned before, I don't have a similar setup over here (only a static ipv6 with prefix delegation at the moment), which makes it a bit difficult to validate solutions on my end. I do want to try to check some assumptions from my end if you don't mind:

-  If I do understand you correctly, the 'Interface bits' in your examples are offered via router advertisements by your provider and thus unknown by any of the interfaces on the system. 
- `2001:db8:1234:*` addresses are only available "downstream" the wan interface is offered a different one (which corresponds with my static setup by the way)

Although I'm still a bit struggling why 'Interface bits'  should be provided separately, I'm highly likely missing something obvious here, maybe I can try to explain how I thought it should/could work and you can shoot holes in my reasoning and get this off the table.  If we need another property, so be it, going back to my main concern, as long as we document it and it's clear for everybody, I don't mind that much, it feels redundant, but if it isn't it isn't.

Let's assume we where offered a network like `2001:db8:1234:/48` in your example and configured three different interfaces internally, for example:

* lan 2001:db8:1234:1::/64
* opt1 2001:db8:1234:2::/64
* opt2 2001:db8:1234:3::/64

My assumption would be that if I would like to describe hosts on for example lan, these would always lie within `2001:db8:1234:1::/64` so a group of hosts (or a single on) could be described as a section within that region (hence the split on /64 in this case and prepending the provided netmask in the alias).

If I do understand you correctly, I'm not aiming for a host within the configured net, but some other host living on another net. I just can't seem to get my head around why that is. Maybe you can help me out here, just assume I don't know anything.

> But as Ad noted, it is essential that we define and document very precisely and without ambiguity what the feature is supposed to do and how to use it.

My proposal would be to first make sure the documentation is good and then change code accordingly if needed, I hope you and @marjohn56 could work on that so we have our "specs in writing first". Sorry to be a bit blunt here, but I truly believe that if we can't describe functionality it's very often a recipe for future disaster, some engineering mistakes are quite hard to fix afterwards (which is the cause for quite some `reply-to` and `route-to` issues for example, as soon as we start "assuming" it's quite difficult to get out of it) .

marjohn56:
In theory a fixed /64 change would work fine for individual hosts and the LAN. I've just re-read https://github.com/opnsense/core/issues/2544#issuecomment-817143665 and now I understand where @maurice-w is coming from, the network alias is for further prefix delegations from the LAN to downstream routers. @fichtner - re: the variable length, the only use I could see for it would flexibility, it just gives the user that option and thus is not forced to use /64, though in 99.9% of cases it would work with a fixed /64.

n-krueger:
Hello, I am interested in this feature.

What work would be necessary to get this functionality documented well enough to be merged? As I understand the discussion, the issue is that it is not clear what the *Dynamic IPv6 Host* alias is supposed to represent.

**/64 subnets of ISP prefix delegation**
My understanding is that this is already implemented by configuring IPv6 on an interface with *Track IPv6 interface* and setting an *IPv6 Prefix ID*. With @AdSchellevis example configuration:
* WAN `2001:db8:1234::/48` prefix delegation
* LAN  tracks WAN with prefix ID 0x1 -> `2001:db8:1234:1::/64`
* OPT1 tracks WAN with prefix ID 0x2 -> `2001:db8:1234:2::/64`
* OPT2 tracks WAN with prefix ID 0x3 -> ``2001:db8:1234:3::/64``

Then the three /64 subnets should be available as *LAN net*, *OPT1 net* and *OPT2 net* in the firewall rules.

**Handling renumbering for specific hosts**
This is one of the use-cases for the proposed *Dynamic IPv6 Host* alias.  The use-case is allowing to create rules for a specific /128 address that is derived from a prefix delegation that changes over time. 

**Specifying subnets of ISP prefix delegation that are not /64**
This is the downstream prefix delegation use-case, where a user might want to specify a derived subnet that is larger than /64, /56 for instance.

It seems a generic solution to this problem would consist of two parts:
1. Creating an alias that tracks the IPv6 subnet of an interface (no suffixing at this step)
2. Creating an alias that suffixes an existing alias.

I would be happy to support with technical writing or coding, if necessary.

AdSchellevis:
@n-krueger its probably best to start with a chapter in the documentation what it should do, I saved the backing code in a separate branch, just don't want to work on it as long as we all can't agree on what it functionally should do (in clear and simple to read text, without hard to understand edge-cases)

bimbar:
> @AdSchellevis
> 
> Interface address for all examples: `2001:db8:1234:5678:4343:65ff:fe56:7878`
> 
> **"IPv6 Dynamic Host(s)":**
> 
> Example a):
> Input 'Content': `::a`
> Input 'Interface bits [1]': `64`
> Output (Alias): `2001:db8:1234:5678::a`
> 
> Example b):
> Input 'Content': `::f:0:0:0:a`
> Input 'Interface bits': `60`
> Output (Alias): `2001:db8:1234:567f::a`
> 
> Example c):
> Input 'Content': `::f:0:0:0:a`
> Input 'Interface bits': `56`
> Output (Alias): `2001:db8:1234:560f::a`
> 
> **"IPv6 Dynamic Network(s)":**
> 
> Example a):
> Input 'Content': `::b/64`
> Input 'Interface bits': `56`
> Output (Alias): `2001:db8:1234:560b::/64`
> 
> Example b):
> Input 'Content': `::c0/60`
> Input 'Interface bits': `56`
> Output (Alias): `2001:db8:1234:56c0::/60`
> 
> Example c):
> Input 'Content': `::ef00/56`
> Input 'Interface bits': `48`
> Output (Alias): `2001:db8:1234:ef00::/56`
> 
> [1] Probably a dropdown, and need to find a better name for "how many bits of the interface's address to use".

I like that approach. Additionally we would need some sort of dropdown which interface to track for the prefix, and I would like to add an additional example:

Example d):
Input 'Track Interface': `LAN`
Input 'Content': `::/64`
Input 'Interface bits': `64`
Output (Alias): `2001:db8:1234:5678::/64`

to match an entire interface network, because the "LAN Net" etc. can not be used as an alias, but only inside the actual firewall rule.

Does that make sense?

fichtner:
The most interesting thing is how we define and merge the prefix, i.e.

The ISP prefix is the first part for example /50.
The Address subnet is the second part, mostly /64.
The gap of bits 51 - 63 is where there the prefix ID lives and needs to be defined also.

The easiest way is to assume the ISP prefix is /64 (tracking interface) and the resulting address is /64 also. It takes care of prefix ID automatically, but doesn't work for prefix delegation since we base the assumption on the active interface address.

n-krueger:
@bimbar @fichtner What do you see as the semantics of "Interface Bits"? To me it seems to duplicate the prefix length information we already have in the interface address. We could let the user specify the size of the subnet they want to create:

Interface Address: `2001:db8:1234:5678:4343:65ff:fe56:7878/56`

**IPv6 Dynamic Host(s):**

Example a):
Input 'Content': ::a
Output (Alias): 2001:db8:1234:5600::a

Example b):
Input 'Content': :0f:0:0:0:a
Output (Alias): 2001:db8:1234:560f:0:0:0:a

Example c):
Input 'Content': :f0:0:0:0:a
Output (Alias): 2001:db8:1234:56f0:0:0:0:a

**IPv6 Dynamic Network(s):**

Example a):
Input 'Content': :0b
Input 'Prefix Length': 64
Output (Alias): 2001:db8:1234:560b::/64

Example b):
Input 'Content': :c
Input 'Prefix Length': 60
Output (Alias): 2001:db8:1234:56c0::/60

Example c):
Input 'Content': :0b:0:0:0:0
Input 'Prefix Length': 127
Output (Alias): 2001:db8:1234:560b:0:0:0:0/127

Example d)
Input 'Content': (disabled)
Input 'Prefix Length': 56
Output (Alias): 2001:db8:1234:56::/56

This way we could also validate that the input for `Content` matches the desired prefix length when combined with the dynamic prefix from the interface.

bimbar:
@n-krueger: Maybe it does.

I think we should fix the example first, and characterize the limitations we have with this approach.

- The interface address we use as an example:
Address: 2001:db8:1234:5678:4343:65ff:fe56:7878/56
Prefix: 2001:db8:1234:5600::/56

- I'm also not sure what interface address that would actually be, since none of them will be /56. You'll probably something like the following (basing off a german fritzbox, mnet as a provider and delegated prefix size /60):
Address: 2001:db8:1234:5601::dead:beef/64
Delegated Prefix: 2001:db8:1234:56e0::/60
LAN Prefix (tracking the WAN IF): 2001:db8:1234:56e0::/64

- The approach can only rewrite prefixes at 4 bit (hex character) boundaries. This might be fine, at least for a start, but prefix delegations can be /62 or something similar.

- I don't really see why we would not treat a host as a /128 network.

So everything I want to do is characterize LAN addresses or the entire LAN network in firewall rules, so it would all be:

Example A1 for a network)
Tracking Interface: LAN
Content: ::/64
Output: 2001:db8:1234:56e0::/64

Example A2 for a host)
Tracking Interface: LAN
Content: ::1/128
Output: 2001:db8:1234:56e0::1/128

which is fairly simple. But what if we want to firewall the entire delegated prefix or subblocks of it:

Example B)
Tracking Interface: ????
Content: 4::/62
Output: 2001:db8:1234:56e4::/62

The Problem here is that I want to track the entire delegated prefix, which is not assigned to any interface, and I can not track WAN, because that in itself is outside the delegated prefix.

Now what if we have a /62 delegated prefix?

Example C1)
Delegated Prefix: 2001:db8:1234:56e4::/60
Tracking Interface: ????
Content: 6::/63
Output: 2001:db8:1234:56e6::/63

Perhaps Content should be an offset like this:

Example C2)
Delegated Prefix: 2001:db8:1234:56e4::/60
Tracking Interface: ????
Content: 2::/63
Output: 2001:db8:1234:56e6::/63

so it also works like this:

Example C3)
Delegated Prefix: 2001:db8:1234:56e8::/60
Tracking Interface: ????
Content: 2::/63
Output: 2001:db8:1234:56ea::/63

That does not even include multiple WAN interfaces and that I'm not sure how "Track Interface" allocates prefixes, so in my opinion B) and C) should not even be implemented.

Thinking about this I would favor a simple solution that works, and just does Example A) and only /64 networks. It would be fine for 99% of cases and comparatively easy to implement. Everything else is full of complexity and unpleasant edge cases we have not thought if.

n-krueger:
@bimbar Fair point on the `/56` prefix likely not being used as-is on an interface.

I agree that hosts can just be /128 subnets. Perhaps we can split the dynamic tracking and subnetting?

**Alias: Prefix Delegation**
Input `Name`: `ISP_PD`
Input `Tracking Interface`: `WAN`
Output: `2001:db8:1234:56e0::/60`

**Alias: Subnet**
Input `Network`: `ISP_PD`
Input `Prefix ID`: `3`
Output: `2001:db8:1234:56e3::/64`

bimbar:
It would be useful to be able to refer to the prefix delegation as well, I agree.

However I think the whole feature is kind of niche and mostly useful to non-commercial users, of which there may not be that many, while commercial users tend to have static IP ranges, so it might not be that crucial of a feature to expend that much effort into it.
Personally I could live perfectly well with just /64 interface tracking.

fichtner:
I agree but I fear the next PD user will ask for this as soon as the feature is in and I don’t want to re-engineer it if it can be avoided by careful consideration. It’s the same topic discussion all over again. ;)

bimbar:
In that case, why not 
Input `Mode: Interface | DHCPv6 PD`

In case of `Mode DHCPv6 PD`
Input `Tracking Interface: WAN`
Input `IPv6 Prefix ID: X`

or in case of `Mode Interface`
Input `Tracking Interface: LAN`

Both of those should reduce it to a prefix, so after that:
Input `Content: ::1/128` or `::/64`.

In case the whole PD is wanted, `IPv6 Prefix ID: 0` and `Content: ::/60` (or whatever the prefix length is) should do it?

The difference to @n-krueger is that I think we have to take into account that the WAN interface gets both a /64 interface address outside of the PD range and for example a /60 PD, so tracking WAN alone without specifying some sort of mode may not work.

marjohn56:
Let's see if we can move this forward, as I have a few hours today and basically, **it's been sat here too long**... the whole thing is turning into something like a navel gazing contest. Looking at my PR for the wiki the alias says 'Dynamic Host Alias'. In this context the key word is 'HOST', the fact that you can specify a variable prefix length is just a bonus; the original thought and work behind this that I and @AdSchellevis did was to allow hosts that were using a dynamic prefix to have the fw dynamically change if the prefix changed, and it does just that, so a 'Dynamic Host Alias' is fixed on a /64 boundary. So we remove the need to enter the prefix length as it will alwatys be a /64.

So we could move this forward by saying we have that, and that is what a 'Dynamic Host Alias' entry does. Now, if we want to go further and add another alias type, such as 'Dynamic Prefix Alias', that becomes a seperate entity. Yes, I realise that the 'Dynamix Prefix Alias' could do the same thing as the 'Dynamic Host Alias' but let's use the kiss principle and allow the less sophisticated user a simple method, the more advanced user may wish to use a 'Dynamic Prefix Alias', but a few lines of code to add another alias type simplifies the wiki too.

This feature is REALLY useful and it works, it's a terrible shame that  it's been sat here for months whilst the Wiki page is discussed. Even more useful to me now as I have moved to an ISP that uses dynamic prefixes, though tbh, that prefix is pretty sticky.

marjohn56:
Please excuse the merge branch, for some reason Github has got it's knickers in a twist. Ignore it. I have just re-read this entire thread, and having read it I better understand where @maurice-w was coming from and also where @AdSchellevis was at. @maurice-w was looking at this ( I think ) of dynamically changing the whole prefix i.e. the /48 or whatever and not bothering per say with the host suffix, whereby @AdSchellevis and I were looking at appending the host suffix to the new prefix. We need to look at this from the usage it's likely to get, which, by and large will be with users who's ISP gives them a dynamix pd, this is unlikely to be pro users as they are most likely going to use static ranges... At the moment I cannot see a reason not to just go with the way it currently works - I've updated the wiki to more correctly show the useage. There may be some edge cases where the user want's to use a different prefix size, but that would appear to be just that, an edge case. If needed, then we can add a ipv6 network alias type, it's pretty much the same. My only question is should it be a /128 that's entered rather than  a /64?

AdSchellevis:
@marjohn56 it has been quite long ago that I looked at this to be honest, but my main concern is probably how the network type should be implemented in terms of user input. if we use the `/64` in the address field as a marker to steer the number of bits to chop, the notation for the network type might be harder to explain (or maybe I'm overlooking something).

If the current documentation is clear for people involved and we can reach consensus on how a network type should look like when being implemented in a later stage, I don't mind adjusting the code where needed to start with a host type.

Copying one of @maurice-w's examples, given the following Interface address `2001:db8:1234:5678:4343:65ff:fe56:7878` on the track interface (LAN), the user would input the following :

Content: `::a/64`
Interface: `LAN`

Which would then generate :  `2001:db8:1234:5678::a` , right?

marjohn56:
Hi @Ad ltnc...

Yes, thats perfectly correct.

marjohn56:
I keep re-reading thread and everytime I read it I come to a different conclusion.  @maurice-w are you still with us on this or have you thrown the towell in?

marjohn56:
OK, I think this is my final take on this.. I hope. 

A host alias should not require and indeed not allow a size when entered. the split will automatically happen on the /64 boundary for that LAN. It is after all intended for a single host.

An IPv6 network alias allows the user to specify the length or bits ( don't like the use of 'bits' as it confuses the issue ), so the user can specify any length they wish, we effectively already have this now. So we are left with, IMHO, a slight adjustment to the validation of the 'HOSTS' entry where it specfically denies a length entry, only allowing the address part, and a new entry for 'Networks' where it specifically requires that you do enter a length, this can either be a manual entry or a drop down, manual entry being my preference.

I do wish to add though that I still see no use for the network alias. The idea is to auto change the fw entry should the prefix on the LAN change; there may be some edge cases where the LAN itself is not /64, for example the ISP only gives out a single /64 pd and the user wishes to break that down, so a network alias possibly comes into play there. Apart from that edge case,  how and where will this IPv6 Network Alias be used? NPtv6 possibly?

AdSchellevis:
@marjohn56 
> A host alias should not require and indeed not allow a size when entered. the split will automatically happen on the /64 boundary for that LAN. It is after all intended for a single host.

The problem is where does the `/64` come from, our (my) previous assumption was the interface (https://github.com/opnsense/core/blob/11ad2a2a6192804f3d392181be21761896545e0d/src/opnsense/scripts/filter/lib/interface.py#L67), after which we ended up that it's not the mask we're seeking. 

So if one should offer it to the system, we need to think about how one should provide it and when thinking a step further, how would that conceptually look like for a network.

> An IPv6 network alias allows the user to specify the length or bits ( don't like the use of 'bits' as it confuses the issue ), so the user can specify any length they wish, we effectively already have this now. So we are left with, IMHO, a slight adjustment to the validation of the 'HOSTS' type entry where it specfically denies a length entry, only allowing the address part, and a new entry for 'Networks' where it specifically requires that you do enter a length, this can either be a manual entry or a drop down, manual entry being my preference.

That would explicitly mean the type of input defined by @maurice-w in https://github.com/opnsense/docs/pull/330#issuecomment-837607090, right? 

The only issue I have with this is that it circles back to my original questions in 

It would still be nice to have some answers about my earlier questions in https://github.com/opnsense/docs/pull/330#issuecomment-839531339 , if we can't know the mask, sure we should ask for it (or if it's to complicated to extract), but it sounds a bit odd. Don't want to make this a breaking concern perse, but for some reason I do have a strong feeling that in a year or so we'll end up in the same loop.

> I do wish to add though that I still see no use for the network alias. The idea is to auto change the fw entry should the prefix on the LAN change; there may be some edge cases where the LAN itself is not /64, for example the ISP only gives out a single /64 pd and the user wishes to break that down, so a network alias possibly comes into play there. Apart from that edge case, how and where will this IPv6 Network Alias be used? NPtv6 possibly?

I would assume this could be practical in firewall or translation rules, but it's totally fine to not offer this functionality until anyone asks as long as we are sure there's room (conceptually) to add it later (hence why I keep hammering on documentation first if we can't seem to identify the correct feature to add).

marjohn56:
Let me try and respond in sections to your comments.

1. 
> > A host alias should not require and indeed not allow a size when entered. the split will automatically happen on the /64 boundary for that LAN. It is after all intended for a single host.
> 
> The problem is where does the `/64` come from, our (my) previous assumption was the interface (https://github.com/opnsense/core/blob/11ad2a2a6192804f3d392181be21761896545e0d/src/opnsense/scripts/filter/lib/interface.py#L67), after which we ended up that it's not the mask we're seeking.

In the case of an IPv6 Host Alias it comes from the LAN. We know that when we use tracking the LAN will always be /64, we really don't even need to find it, how about we call it an 'Ipv6 tracking host alias', then it's clear what it's intended for.

marjohn56:
2.
> > An IPv6 network alias allows the user to specify the length or bits ( don't like the use of 'bits' as it confuses the issue ), so the user can specify any length they wish, we effectively already have this now. So we are left with, IMHO, a slight adjustment to the validation of the 'HOSTS' type entry where it specfically denies a length entry, only allowing the address part, and a new entry for 'Networks' where it specifically requires that you do enter a length, this can either be a manual entry or a drop down, manual entry being my preference.
> 
> That would explicitly mean the type of input defined by @maurice-w in [#330 (comment)](https://github.com/opnsense/docs/pull/330#issuecomment-837607090), right?
> 
> The only issue I have with this is that it circles back to my original questions in
> 
> It would still be nice to have some answers about my earlier questions in [#330 (comment)](https://github.com/opnsense/docs/pull/330#issuecomment-839531339) , if we can't know the mask, sure we should ask for it (or if it's to complicated to extract), but it sounds a bit odd. Don't want to make this a breaking concern perse, but for some reason I do have a strong feeling that in a year or so we'll end up in the same loop.
> 
> > I do wish to add though that I still see no use for the network alias. The idea is to auto change the fw entry should the prefix on the LAN change; there may be some edge cases where the LAN itself is not /64, for example the ISP only gives out a single /64 pd and the user wishes to break that down, so a network alias possibly comes into play there. Apart from that edge case, how and where will this IPv6 Network Alias be used? NPtv6 possibly?
> 
> I would assume this could be practical in firewall or translation rules, but it's totally fine to not offer this functionality until anyone asks as long as we are sure there's room (conceptually) to add it later (hence why I keep hammering on documentation first if we can't seem to identify the correct feature to add).

I think for sanitys sake we need to break this down, a 'one alias for them all' really does not cut the mustard. We can add further aliases when asked for and when clearly defined by the requestee. We know exactly what we want to do with the Ipv6 host alias or I believe we do. It is to allow a host(s) where a firewall rule is needed, such as a web server, to automatically track a changed pd.

marjohn56:
I would like @maurice-w to come back in, but he seems to have left the planet. Either way, we need more than one alias type.

AdSchellevis:
@marjohn56 

> In the case of an IPv6 Host Alias it comes from the LAN. We know that when we use tracking the LAN will always be /64, we really don't even need to find it, how about we call it an 'Ipv6 tracking host alias', then it's clear what it's intended for.

Naming the type is the last of my concerns, as long as the intend is clear.  If the address and the mask both come from lan, I'm still probably missing something. But if I'm the only one doubting and we really think we should provide the netmask (/64) separately, maybe we should for the sake of progress add a one. I will just feel very stupid if at some point in time it turns out the mask was known all along. (I don't have access to such a setup, so I can only work with what you're providing here)

> I think for sanities sake we need to break this down, a 'one alias for them all' really does not cut the mustard. We can add further aliases when asked for and when clearly defined by the requestee. We know exactly what we want to do with the Ipv6 host alias or I believe we do. It is to allow a host(s) where a firewall rule is needed, such as a web server, to automatically track a changed pd.

I wasn't proposing a single alias for all, just making sure that the next alias will look similar for apparent reasons (e.g. a host and network alias are also the same, just leaving out the netmask, as one would expect).

marjohn56:
It's really easy to set up a system so you can see what happens. Surely you have access to some hardware! Try installing Opnsense I understand it's very good.😊

It's straight forward really. We have a LAN, which has a pd let's assume the pd length is /56 and our lan we know has the prefix ID 0, that's our tracking ID. Our GUA prefix given by the ISP is 2a00:1111:1111:00/56. We have a server on the LAN, which is a a /64 and the server address, lower 64 bits only is 0000:0000:0000:0001. At this moment, the full GUA address of the server is:

LAN 64 from prefix        our entry for our alias
2a00:1111:1111:0000:    0000:0000:0000:0001 

and for pedants sake, call it a /128 and we have our ailias firewall entry.

Now, our wonderful ISP changes our prefix to: 2a00:1111:1111:01/56

That triggers our routine... we only really need to change the upper 56 bits, but we would then have to go find the pd length and start working that out,  !!!BUT!!!, we can look at what the Routers LAN address is and as we know that our LAN is 64 bits, we can take the upper 64 bits of that address in its entirety and use that. so now our new firewall entry address is:
Lan 64 from prefix             our entry for our alias
2a00:1111:1111:0100:       0000:0000:0000:0001

Does that make it any clearer?

AdSchellevis:
> It's really easy to set up a system so you can see what happens. Surely you have access to some hardware! Try installing Opnsense I understand it's very good.😊

You're being too funny.

> That triggers our routine... we only really need to change the upper 56 bits, but we would then have to go find the pd length and start working that out, !!!BUT!!!, we can look at what the Routers LAN address is and as we know that our LAN is 64 bits, we can take the upper 64 bits of that address in its entirety and use that. so now our new firewall entry address is:
> Lan 64 from prefix || our entry for our alias
> 2a00:1111:1111:0100: || 0000:0000:0000:0001
> 
> Does that make it any clearer?

Yes and no, the way you explain it, our current approach is what we're looking for (mask is defined by the interface), reading back the noise started when we needed to input that mask.

> Example a):
> Input 'Content': ::a
> Input 'Interface bits [1]': 64
> Output (Alias): 2001:db8:1234:5678::a

So probably my only question is, the 'Interface bits [1]' in the example above is that the number found on the tracking interface and thus configured in what we already build or are we talking about something different here?

marjohn56:
Yes, this is where 'bits[1]' starts to get confusing. Forget bits[1]....

Our address length in total is 128 bits. The LAN is always going to be a /64 - the lower 64 bits, at least where we are talking tracking. Yes, there are some edge cases but we they are just that, edge cases and really the ISPs should be put to death.

So taking the example above
Input 'Content': ::a = lower 64 or :0000:0000:0000:0000a
Input 'Interface bits [1]': 64 = upper 64 or 2001:0db8:0000:5678:

Output (Alias): 2001:0db8:0000:5678:0000:0000:0000:0000a or 2001:db8:0:5678::a

Make sense?

marjohn56:
We CAN enter the mask, but the mask is really only for the rule. We will only ever break the addresses on the /64 boundary and adjust the alias, thus a /128 for an alias rule would be the lower 64 bits, in the case above the content ::a/128, so after we've jouned on the /64 boundary our rule entry is 2001:db8:0:5678::a/128

AdSchellevis:
> We CAN enter the mask, but the mask is really only for the rule. We will only ever break the addresses on the /64 boundary and adjust the alias, thus a /128 for an alias rule would be the lower 64 bits, in the case above the content ::a/128, so after we've jouned on the /64 boundary our rule becomes is 2001:db8:0:5678::a/128

That's about how I understood it the first time, we'll fetch the `/64` from the interface and us it to zero the lower 64 bits next merge the user input resulting in the host entry (which would be a /128 when talking about a single host). I guess that's also what the current code does. 

But then @maurice-w mentioned :

> [1] Probably a dropdown, and need to find a better name for "how many bits of the interface's address to use".

... and then I'm lost, it's either defined by the interface or it isn't, so what's your take on that? If we can't explain why one should enter the mask, I think we shouldn't add an extra dropdown (we can add it later if someone can explain why that is)

marjohn56:
Methinks this is for his network alias. As we are going to create a separate Network Alias, when and if someone describes the operation exactly, then I think we can stick with what we have for the host alias. If you are in agreement then I will expand on the wiki and we can move forward with the host alias as it stands.

AdSchellevis:
@marjohn56 if you can finish the docs, maybe add some of the examples out of this thread as well, I'll try to dust of what we have, finish the validations (no mask allowed, only an address) and we're good to go. Thanks!

marjohn56:
OK, I've done that. I'm going to create a new PR for it. We'll leave this one as a discussion document for network alias if and when.

AdSchellevis:
@marjohn56 with the code in place and the docs waiting for a release, shall we close this one now?

marjohn56:
I Think we can close this now.. There will I expect be further discussions on the network alias but we have decided that's a different beast..

PR #335, opened 2021-05-07
Virtual IPs menu was switched from Firewall to Interfaces

ypid:
Don’t know the exact version this was done of the tip of my head. At least it is true for OPNsense 21.1.5.


AdSchellevis:
@ypid thanks, it has been this while for quite some time, apparently we totally missed it in the docs.

PR #344, opened 2021-06-19
Rewrite of WireGuard client how-to

Greelan:
Given some comments I've seen recently on the forum and Reddit from users who are confused by the existing how-to, I am suggesting a re-write to make it clearer and more step-by-step, while also adopting a similar structure to the selective routing how-to

### DIFF
--- a/source/manual/how-tos/wireguard-client.rst
+++ b/source/manual/how-tos/wireguard-client.rst
@@ -67,11 +67,12 @@ Step 3 - Configure the endpoint (client peer)
 - **Save** the Local peer configuration again, and then click **Save** once more
 - Repeat this Step 3 for as many clients as you wish to configure
 
---------------------------
-Step 4 - Turn on WireGuard
---------------------------
+----------------------------------
+Step 4 - Turn on/restart WireGuard
+----------------------------------
 
-Turn on WireGuard under :menuselection:`VPN --> WireGuard --> General` if it is not already on
+- Turn on WireGuard under :menuselection:`VPN --> WireGuard --> General` if it is not already on (click **Apply** after checking the checkbox)
+- Otherwise, restart WireGuard - you can do this by turning it off and on under :menuselection:`VPN --> WireGuard --> General` (click **Apply** after both unchecking and checking the checkbox)
 
 --------------------------------
 Step 5 - Assignments and routing
@@ -79,7 +80,7 @@ Step 5 - Assignments and routing
 
 .. Note::
 
-    The steps outlined in Steps 5(a) and 5(b) below may not be required at all in your circumstances. Strictly speaking, if you only intend for your clients to use the tunnel to access local IPs/subnets behind OPNsense, then neither step is actually necessary. If you intend to use the WireGuard tunnel to also access IPs outside of the local network, for example the public internet, then at least one, and perhaps both, of the steps will be required
+    The steps outlined in Steps 5(a) and 5(b) below may not be required at all in your circumstances. Strictly speaking, if you only intend for your clients to use the tunnel to access local IPs/subnets behind OPNsense, then neither step is actually necessary. If you intend to use the WireGuard tunnel to also access IPs outside of the local network, for example the public internet, then at least one, and perhaps both, of the steps will be required. This is explained below
 
     **However**, it is useful to complete Step 5(a) anyway, for the reasons explained in that step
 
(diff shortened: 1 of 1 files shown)


mimugmail:
Hm, I'm a bit confused. But not your fault, it was mine. This article is about configuring OPN as Central server, but everything after assigning interface is only relevant If OPN is the client (like in mullvad), which might be the main reason why ppl get confused. Again, not your fault since it was already there before. 

What do you think about a basic guide, like yours but stopping before assigning interface, then a second one convering advanced examples with you basic article as a prerequirement?

Greelan:
@mimugmail thanks for the feedback. It is quite possible that I am confused, not you. 🤯 My understanding, based not only on the existing how-to but other guides I have seen, is that an interface is required if the client is intended to access the internet via the OPNsense server, as opposed to just local networks. To be honest, I haven't actually tested to confirm that is the case (I will if I get some time). 

Maybe the real issue is the outbound NAT rule - adding the interface automatically adds the interface subnets to the automatic outbound NAT rule, which is perhaps the reason why some have considered the interface necessary. So the solution may be either (a) just add an interface and don't add a manual outbound NAT rule, or (b) don't add an interface but just create a manual outbound NAT rule. Option (a) won't solve for IPv6 though if required to NAT ULAs.

But in any event, as I note in my re-write, assigning an interface is actually useful given it generates the "net" alias for the interface that can then be used in firewall rules. And it means that firewall rules can be organised by wgX device, rather than putting them all together in the WireGuard group.

If it can be verified that an interface is not required for the client to access non-local networks, I can make this clearer in the re-write? And point out that if an interface is not created, manual outbound NAT rules are required?

Greelan:
OK, testing shows it all comes down to outbound NAT.

1. With interface assigned:

Firewall rules are included on the interface, using the "net" alias as source

(a) No manual outbound NAT rules:
  - IPv4 can access the internet (because the interface is included in the automatic outbound NAT rule)
  - IPv6 cannot access the internet (because the automatic outbound NAT rule applies only to IPv4, add I have ULAs on my tunnel, not GUAs)

(b) With manual outbound NAT rules for IPv4 and IPv6:
  - both can access the internet, although given (a) only IPv6 needs the manual rule

2. No interface assigned:

Firewall rules are included on the WireGuard group, using a manually defined alias as source

(a) No manual outbound NAT rules:
  - neither IPv4 nor IPv6 (given ULAs are configured, not GUAs) can access the internet

(b) With manual outbound NAT rules for IPv4 and IPv6:
  - both can access the internet

Stepping back from it all, my view is that adding an interface is the cleanest and simplest option, because:
- for IPv4 outbound NAT is automatically dealt with
- it provides a separate place for rules for each wgX device
- it provides an automatic "net" alias that can be used in firewall rules, without the need to define a manual alias
- only if IPv6 ULAs are used is an additional manual outbound NAT rule required

Thoughts?

Greelan:
BTW, the confusion I have seen people expressing with the existing how-to is not just regarding the interface assignment. It is also around AllowedIPs on both OPNsense and client; firewall rules; why a port forward is stated to be necessary; and which keys go on what. Hence I thought a more detailed and step-by-step guide, spelling everything out explicitly, would be a good idea

Greelan:
See what you think of my latest changes, to give more colour on assignments and routing and reflecting my testing above

source/manual/how-tos/wireguard-client.rst
```diff
@@ -6,179 +6,261 @@ WireGuard Road Warrior Setup
...
+- In the dropdown next to “New interface:”, select the WireGuard device (:code:`wg0` if this is your first one)
+- Add a description (eg :code:`HomeWireGuard`)
+- Click **+** to add it, then click **Save**
+- Then select your new interface under the Interfaces menu
+- Configure it as follows (if an option is not mentioned below, leave it as the default):
 
-If more granular rules are required note there is a new interface **wg0** where these may be configured.
+    ============================= ===================================================================
+     **Enable**                    *Checked*
+     **Lock**                      *Checked if you wish to*
```
mimugmail, line 106: Lock is a must when you assign interfaces which dont exist in startup (usually all VPNs) otherwise your interfaces get scrambled

mimugmail (approved):
Really nice work here, love it. Just the one thing with Interface lock

source/manual/how-tos/wireguard-client.rst
Greelan: Thanks. Now fixed.

AdSchellevis:
@Greelan @mimugmail ready to merge?

Greelan:
@AdSchellevis all good from my perspective.

AdSchellevis:
@Greelan thanks!

PR #348, opened 2021-08-07
Update DNSBL Cron Command

Napsterbater:
“Download Unbound DNSBLs and restart” -> "Update Unbound DNSBLs"

“Download Unbound DNSBLs and restart” is no longer an entry, it is now simply "Update Unbound DNSBLs"


AdSchellevis:
@Napsterbater thanks!

PR #353, opened 2021-09-19
WG road warrior/selective routing updates

Greelan:
Adding a specific warning regarding setting DNS servers in the WG road warrior how-to. Also fixing some formatting issues.

Adding instructions for IPv6 configuration in the WG selective routing how-to. Also fixing some formatting issues, and adding some other minor clarifications/explanations.


fichtner:
Merged, thanks!

PR #359, opened 2021-11-01
remove peer id from phase 1

arkste:
Peer ID got removed for Mobile Client's IPSec Phase 1 Configuration years ago (see [0dd120f](https://github.com/opnsense/core/commit/0dd120ff1aac446f0df9a3f82051b95b64eecf27)), because the value is not used in the ipsec configuration.


AdSchellevis:
@arkste thanks!

PR #362, opened 2021-11-15
add more info what actually happens when nut is enabled

tessus:
/ref https://forum.opnsense.org/index.php?topic=25581.0


mimugmail (approved):
Great, thx

tessus:
Is there something still missing? Just asking because it has been approved but not merged yet.

fichtner:
Merged, thanks!

PR #365, opened 2021-11-20
Update wireguard-selective-routing.rst

Greelan:
Updated to reflect discussion in https://github.com/opnsense/core/issues/5329.  Also added instructions for "kill switch".

### DIFF
--- a/source/manual/how-tos/wireguard-selective-routing.rst
+++ b/source/manual/how-tos/wireguard-selective-routing.rst
@@ -240,9 +240,9 @@ Step 10 - Create an outbound NAT rule
 
 - **Save** the rule, and then click **Apply changes**
 
------------------------------------------
-Step 10 - Adding a kill switch (optional)
------------------------------------------
+--------------------------------------
+Step 10 - Add a kill switch (optional)
+--------------------------------------
 
 If the VPN tunnel gateway goes offline, then traffic intended for the VPN may go out the normal WAN gateway. There are a couple of ways to avoid this, one of which is outlined here:
 
--- a/source/manual/how-tos/wireguard-selective-routing.rst
+++ b/source/manual/how-tos/wireguard-selective-routing.rst
@@ -87,8 +87,9 @@ Step 4 - Assign an interface to WireGuard and enable it
      **Enable**                    *Checked*
      **Lock**                      *Checked if you wish to*
      **Description**               *Same as under Assignments, if this box is not already populated*
-     **IPv4 Configuration Type**   *None*
-     **IPv6 Configuration Type**   *None*
+     **IPv4 Configuration Type**   *Static IPv4*
+     **IPv4 Address**              *As per the Tunnel Address under the WireGuard local peer configuration*
+     **IPv4 Upstream Gateway**     *Auto-detect*
     ============================= ===================================================================
 
 - **Save** the interface configuration and then click **Apply changes**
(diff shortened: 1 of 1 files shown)


source/manual/how-tos/wireguard-selective-routing.rst
```diff
@@ -87,8 +87,9 @@ Step 4 - Assign an interface to WireGuard and enable it
      **Enable**                    *Checked*
      **Lock**                      *Checked if you wish to*
      **Description**               *Same as under Assignments, if this box is not already populated*
-     **IPv4 Configuration Type**   *None*
-     **IPv6 Configuration Type**   *None*
+     **IPv4 Configuration Type**   *Static IPv4*
+     **IPv4 Address**              *As per the Tunnel Address under the WireGuard local peer configuration*
+     **IPv4 Upstream Gateway**     *Auto-detect*
```
AdSchellevis, line 92: @Greelan although I don't want to rewind earlier discussions, but are you sure "auto-detect" is enough here. I would expect it needs either an explicit gateway (since no router file present) or an explicit rule (possible with https://github.com/opnsense/core/commit/b567ddb1f1883dc4b14838fe4b06f40709d223b5), where the latter might be cleaner as it doesn't require changes to the interface.

Greelan: @AdSchellevis fair question, I can't test directly as I don't run a local DNS resolver on OPNsense. Maybe @schnerring can test. 

To explain my thinking though:

- the help text for the upstream gateway setting on the Interface configuration says: "For single WAN interfaces a gateway must be created but set to auto-detect."

- the gateway is otherwise specifically assigned to the interface under the Gateway configuration 

- the dropdown for the gateway in the Interface configuration has only the assigned gateway listed, implying that is the one that is auto-detected

Is that reasonable? Of course I am happy to change if needed

AdSchellevis: @Greelan I see your point, but given wireguard doesn't create a `_router` file, it wouldn't add a gateway automatically either (https://github.com/opnsense/core/blob/f68816a8c2f2f02852c4846c0b6806a0fdf42c2e/src/opnsense/mvc/app/library/OPNsense/Routing/Gateways.php#L242-L249)

At some point in time the gateway selection would just "choose one" from the gateway list, which is quite undesirable as well, which means that if there's a clear upstream it might use it (dhcp generates a file, when dynamic gateway is set it also unambiguous). This is likely the reason for discussions in the core ticket as well. 

From a configuration perspective, in my humble opinion it would make most sense to explicitly define an outbound rule with gateway selection (required https://github.com/opnsense/core/commit/b567ddb1f1883dc4b14838fe4b06f40709d223b5), as a "dynamic interface" without an explicit address set might not like being reset with a new address in some situations either.

I hope this makes some sense, as I don't want to cause more noise.

Greelan: @AdSchellevis no problem, it is good to have the expert technical input. I don't profess to fully understand the nuances here 😅 

So to be clear, you are talking about a rule like that suggested here: https://github.com/opnsense/core/issues/5329#issuecomment-968207926 (configured either as a floating rule or, with the patch you mentioned, as an interface rule)?

schnerring: Do I `opnsense-patch b567ddb` and create an outbound rule to test this?

Greelan: The patch isn't necessary to test the "auto-detect" functionality on the interface config. But as per @AdSchellevis' later comments, this approach may be changed to just use an outbound FW rule. The patch is needed if that rule is configured on the interface, rather than as a Floating rule

AdSchellevis: @schnerring to remove the limitation you can use patch, which should by the way have the same affect as a floating quick rule bound to the same interface as @Greelan mentioned. 

@Greelan the outbound rule from @kulikov-a looks similar as the "auto-generated" ones when there's a "router" file indeed.

Greelan: Just to clarify: the rule suggested by @kulikov-a was a **non-quick** Floating rule. Which is correct? And if the rule is configured on the interface itself (with the patch applied), should it be quick or non-quick?

AdSchellevis: @Greelan my mistake, both work, the user defined non-quick rule will overrule autogenerated ones when there, so in this case I would expect non-quick and quick to have the same result, but since the autogenerated was non quick, best to keep this non-quick too. sorry for the noise.

kulikov-a: @Greelan Yes, I also think that the both variants are working. just one more thing: I tried to mimic the syntax of the system "auto-generated" rules. but did not include the `allow-opts` flag. so if full repetition of the system rulea is required, then you should enable "allow options" in the rule

Greelan: See what you think of the changes

schnerring: @Greelan If general kill switch behavior is desired, wouldn't a simpler alternative for kill switch behavior be to check the **Skip rules** option under Firewall &rarr; Settings &rarr; Advanced?

Greelan: That is an alternative (I mention in the how-to there are more ways than one). But not sure it is simpler - a block floating/interface rule still needs to be created for the default gateway for the WAN interface. And anyway I thought that it might interfere with behaviour that the user wants for other purposes, so figured a more tailored solution might be better

schnerring: I tried getting this to work for a couple of hours. I can confirm that the floating rule generates the correct rule. But as soon as I remove the static configuration from the WG interface, things stop working, at least for Unbound and the **Outgoing Interfaces** option.

schnerring (approved):

Greelan:
Hoping that this can be merged now given it reflects the preferred solution @AdSchellevis ?

mimugmail (approved):

AdSchellevis:
@Greelan thanks! I kind of forgot about this one

PR #366, opened 2021-11-24
Add Documentation from #3185

gratuxri:
For tests I have used vyos with pppoe-server: https://docs.vyos.io/en/latest/configuration/service/pppoe-server.html


gratuxri:
https://github.com/opnsense/core/issues/3185

AdSchellevis:
I don't think we should add only a script to the docs,  different options(https://github.com/opnsense/core/issues/3185#issuecomment-589240675)  are probably still on the table, but if we're going for a documentation solution there should be at least some background about what the intended fix would offer and it should fit documentation style (code examples wrapped in a code block, preferably in a separate page since it's only for a selected audience).

gratuxri:
Sorry, it looks better now. Should I open parallel new Pull Request for core? And here explain more about this fix and create extra page?

AdSchellevis:
@gratuxri I was looking for a bit more context, but since you're using my previous script, I've decided to stack it for release with an option in our HA settings (https://github.com/opnsense/core/commit/2cfd70b7ebbcfc9058542b58466176ecda217209). If you can spare some time to write a bit of documentation about the new option, that would be appreciated.

PR #374, opened 2022-01-17
Migrations: note the parent method call necessity

kulikov-a:
Hi!
for nginx migration fix discussion.
Perhaps it is worth paying additional attention to the need to call the parent method?
Thanks!


AdSchellevis:
@kulikov-a better to mention indeed, although it is common practice from an OOP perspective (inherit behaviour and extend)

kulikov-a:
@AdSchellevis Thanks! I agree about the OOP principles (just for some reason I first decided that the parent function call was laid down in `run_migrations.php` itself. maybe I'm not the only one) )

PR #379, opened 2022-02-17
Fixed Step 8 of manual/how-tos/wireguard-selective-routing: Destination / Invert | Unchecked

jonaskonig:
After wondering about why my wireguard routing was not working, I figured out, that in Step 8 of manual/how-tos/wireguard-selective-routing in the residing Interface rule it states:
Destination / Invert | Checked
Wich does not work. It has to be uncheck, thats what I changed.


Greelan (commented):
This is not the solution to the concern identified.

As explained in the how-to, the purpose of this rule is to *exclude* relevant local IPs from using the tunnel. This is aimed at ensuring that the relevant hosts can still access local resources, whether DNS, file storage, local websites or whatever.

The assumption is that no local IPs need to be accessed over the tunnel. However, that might not be true in all cases, eg if some local IPs need to be accessed at the VPN endpoint. In that case, the solution is simply to define an alias that, rather than specifying all RFC1918 IPs, only specifies those that need to be excluded from the tunnel. Then that alias would be used in the rule, with the destination still inverted.

I will prepare changes to the how-to to make this even more explicit.

PR #380, opened 2022-02-18
WG selective routing how-to clarifications

Greelan:
Clarification regarding the exclusion of local networks/IPs from the tunnel as needed


FingerlessGlov3s:
Why the change from RFC1918_Networks to Local_Resources?

FingerlessGlov3s:
I also have that requirement, I create a firewall rule to allow the destination IP, and tell it to use the VPN gateway. Which is essentially policy based routing, which solves that issue.

I find its easier to maintain, than chopping up the RFC1918 networks list.

Thoughts?

Greelan:
That's also an option, and in fact is mentioned at the end of the how-to in the context of DNS leaks. It's also legitimate just to limit the step 8 firewall rule to what is strictly needed - which I do by using the "net" alias for my local networks firewall group as the inverted destination. Again, the point is to get people to think about what they do and don't want to exclude, otherwise as demonstrated [here](https://github.com/opnsense/docs/pull/379) they may just follow the how-to verbatim and wonder why it is not routing as expected

FingerlessGlov3s:
Indeed both ways work but I personally think using RFC1918 is just simpler. Both ways work, but at least using RFC1918 will auto cover any other networks they create. With a security hat on, I wouldn't want to route any RFC1918 networks over the tunnel unless required, so its internet traffic only, then if there's a internal IP on the VPN I need to get too I add the rule in.

You should be using INVERT RFC1918 destination on your normal WAN as well. If a network requires internet access, you should only allow traffic to the internet. Why wouldn't you apply the same logic to the VPN tunnel?

If a device tries to get to a local IP that doesn't exist locally, you shouldn't then try route it over the WAN or VPN. The firewall should be just blocking it. If that IP is on another network, on the other side of some form of tunnel, there should be a rule to allow it. I don't think blindly allowing local networks over the tunnel is a good idea.

Greelan:
I am not at all wedded to these changes. I was trying to give more explanation of what is going on to the other poster and anyone else using the how-to. If you would prefer to present the alternative changes, go for it.

FingerlessGlov3s:
If DNS via the tunnel is also required to stop DNS leak, the best solution is to tell the client to use 1.1.1.1 or 9.9.9.9. Sometimes using the VPN providers provided DNS is in recursive mode, which can actually causes issues loading websites like Netflix, because they now make it so their DNS name servers don't return the correct records if coming from a known VPN IP. If you still use a public resolver you can access Netflix fine using the VPN IP.

If they want to use the VPN providers DNS, then give them a slightly more complicated guide.

mimugmail (approved):
To me it looks good :) But I think one of you guys wrote this sections so I leave this up to you.

AdSchellevis:
@Greelan thanks!

PR #382, opened 2022-02-27
Add Gitlab to the URL format hint

thorian93:
See https://github.com/opnsense/plugins/issues/2866 for reference.

### DIFF
--- a/source/manual/git-backup.rst
+++ b/source/manual/git-backup.rst
@@ -89,7 +89,7 @@ your git server.
 
 .. Tip::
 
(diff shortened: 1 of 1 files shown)


AdSchellevis:
@thorian93 mind capitalising the github phrase as well? looks a bit nicer if both are written similar (GitHub, GitLab) ;)

thorian93:
I actually thought about that, and then I thought: 'No, the person writing that line will have put thought into the writing. ;)
Also, is it really 'GitHub'? I have seen several forms of writing. But I am gonna change it, as it feels prettier anyway.

AdSchellevis:
thanks! It seems to be GitHub indeed according to their frontpage title

PR #387, opened 2022-03-11
Update Zen UK IPv6 guide to use DHCPv6 for WAN

nirojan:
Recently IPv6 stopped working for me on Zen UK connection and on reaching out to their support, was informed that they are phasing out support for static IPv6 and advised switching to DHCPv6 for WAN interface.

I have amended the guide to reflect this change. Also, added a note about it being deprecated as an information for users who already have static WAN setup and happen to stumble upon the guide again.


marjohn56:
As I no longer use Zen then we'll have to take what @nirojan as fact. I have no objections.

AdSchellevis:
@nirojan sounds good, thanks!

PR #391, opened 2022-04-02
Update NAT64 interface address documentation

maurice-w:
As of OPNsense 22.1, NAT64 interface addresses must not be used by another interface or VIP.

Reported on the forum by user cugerm and reproduced by me:
https://forum.opnsense.org/index.php?topic=27511


AdSchellevis:
@maurice-w thanks!

PR #401, opened 2022-05-11
Added hint for "client specific overrides"

JochenKorge:
I totally tripped over those "missing" lines.
I know it is said in the manual, but most google matches will lead directly to the how-to. 
It took me way to long to find the "bug" so this little Hint should prevent others from the same mistake.

Cheers Jochen


AdSchellevis:
fair enough, repeating this wouldn't hurt

PR #411, opened 2022-06-23
Additional info about Virtual IPs of type Other

Emi94:
This sentence from pfsense docs helped me realize I should use type Other instead of IP Alias for a subnet that's routed to my WAN IP address.


AdSchellevis:
@Emi94 thanks! Although aliases are more flexible for these use-cases, I don't mind adding the options here as well.

PR #412, opened 2022-06-26
changed some services from GET to POST

lisn0:
changed ids start, stop, restart methods from GET to POST


AdSchellevis:
these are auto-generated by https://github.com/opnsense/docs/blob/master/collect_api_endpoints.py, adjusting the page doesn't work in this case.

PR #420, opened 2022-08-04
Explain automatic nightly remote backup in cloud_backup.rst

mzurhorst:
Explain the automatic nightly remote backup in more detail, and also explain that additional Cronjobs will only be required when either more frequent backups or backups at another time of the day would be required.

Per discussion in:  https://github.com/opnsense/core/issues/5931


mzurhorst:
I am wondering whether there would be anything wrong with this PR that it cannot get merged into the documentation?

AdSchellevis:
@mzurhorst certainly not, just slipped our attention. thanks!

PR #424, opened 2022-08-18
add some captions to HelloWorld example

KDGundermann:


AdSchellevis:
@KDGundermann thanks!

Issue #423, opened 2022-08-18
add some captions to the HelloWorld documentation 


KDGundermann:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [X] I have read the contributing guide lines at https://github.com/opnsense/docs/blob/master/CONTRIBUTING.md
- [X] I have searched the existing issues and I am convinced that mine is new.

**Is your feature request related to a problem? Please describe.**

add some captions to the HelloWorld documentation

AdSchellevis:
https://github.com/opnsense/docs/pull/426

PR #426, opened 2022-08-18
#423:add directory structure for HelloWorld example

KDGundermann:


AdSchellevis:
@KDGundermann thanks! looks like a nice addition.

PR #434, opened 2022-10-22
Fix grammar

asmolero:


source/manual/software_included.rst
```diff
@@ -72,7 +72,7 @@ If, for example you would like to install the gnu `nano <https://www.nano-editor
 The ports tree
 ....................................
 
-In case your using software, which is not supplied by us, you can always build these packages yourself.
+In case you're using software, which is not supplied by us, you can always build these packages yourself.
```
fichtner, line 75:
```suggestion
In case you are using software, which is not supplied by us, you can always build these packages yourself.
```

fichtner:
(Contractions should be avoided in technical writings)

PR #438, opened 2022-11-22
Add privileges/authorization info to the API

wrobelda:
This adds information about the required privileges to the API pages. Uses the corresponding ACL.xml, if found in the source code.

### DIFF
--- a/collect_api_endpoints.in
+++ b/collect_api_endpoints.in
@@ -2,10 +2,10 @@
 {{ title_underline }}
 {% for controller in controllers %}
 .. csv-table:: {{controller.type}} ({{controller.filename}})
-   :header: "Method", "Module", "Controller", "Command", "Parameters"
-   :widths: 4, 15, 15, 30, 40
+   :header: "Method", "Module", "Controller", "Command", "Parameters", "Privilege required"
+   :widths: 4, 15, 15, 30, 40, 40
 {%     for endpoint in controller.endpoints %}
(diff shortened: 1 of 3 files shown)


wrobelda:
FYI, `DEFAULT_BASE_METHODS` need some fixing, they aren't getting their ACL generated. I'll ping once ready.

wrobelda:
Done.

AdSchellevis:
It's certainly a good idea to try to add the ACL information, but it's a bit more complicated to merge this into the scripts, ACL's could be part of the core repo as well (and plugins are allowed to have different models). I'll try to take a look as soon as I can find some time.

wrobelda:
> It's certainly a good idea to try to add the ACL information, but it's a bit more complicated to merge this into the scripts, ACL's could be part of the core repo as well (and plugins are allowed to have different models). I'll try to take a look as soon as I can find some time.

I suppose this can't harm for the time being? And while plugins can have different models, correct me if I am wrong, but the `collect_api_endpoints.py` assumes one specific structure, which I am building upon?

AdSchellevis:
The script certainly won't capture everything, but when it comes to ACL's and menu registrations the reality is a bit more complicated (see for example the ACL's in https://github.com/opnsense/core/blob/master/src/opnsense/mvc/app/models/OPNsense/Core/ACL/ACL.xml). The same endpoint may also exist in multiple ACL's. I haven't looked in depth to your code yet, but before we are adding this, we should at least make sure this works for 90% of core's MVC features to prevent future discussions.

So, as soon as I do have some time available, I'll take a look. By the way, the user/group manager also shows which endpoints a role (privilege) contains.
image

PR #439, opened 2022-11-28
Fix spelling

biddellns:
dounting -> daunting

I googled online and couldn't find a source for dounting so I updated the spelling issue here. If this still happens to be a case of British vs American spelling, feel free to close this PR. 

Cheers!


AdSchellevis:
@biddellns thanks, looks like a typo indeed

PR #442, opened 2022-12-12
Update two_factor.rst - including note that this impacts sudo

alexwasserman:
Adding a note to the "usage" section that this will impact sudo usage too


AdSchellevis:
I'm not sure this will add much to be honest,  the same is true for any other service that requires a password (and uses the authenticator in question), such as re-authentication of vpn packets, sudo and likely others. Rephrasing might be an option if for some reason it's not clear that the password question is never about password alone in these cases.

fichtner:
Plus this is only true when integrated authentication setting is turned on.

fichtner:
I'm going to close this after no refining of the concerns were proposed.

PR #443, opened 2022-12-13
docs: fixed documentation related to GDrive backup

bgdanix:
The documentation about setting up backups into Google Drive is incorrect as it asks you to enter the Service Account email address, while the correct way to set it is to use the unique ID of the SA. Otherwise, an error like below is thrown:

```shell
"error":"invalid_grant","error_description":"Invalid grant: account not found"
```


bgdanix:
@AdSchellevis maybe you can take a look

PR #444, opened 2022-12-18
Some warnings and an example for clarity

TeamLinux01:
Added a warning about overlapping networks for the LAN and Wireguard interfaces, as doing so will cause the clients not to be able to communicate with each other properly. Added an example to request another PD from your ISP, as ATT only hands out /64s and one will not be enough to subnet the networks properly.

### DIFF
--- a/source/manual/how-tos/wireguard-client.rst
+++ b/source/manual/how-tos/wireguard-client.rst
@@ -38,7 +38,38 @@ Step 2 - Configure the local peer (server)
 
 .. Note::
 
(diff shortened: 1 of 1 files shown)


Greelan (changes requested):
The how-to already notes that the tunnel IP and subnet must be *unique*, so the suggested warning is just repeating that. The warning is also too narrow - the tunnel subnet must be distinct from *any* other subnet configured on OPNsense, whether LAN or any VLAN.

I also don’t think this is the right place for the example script. It clutters the how-to with detail for a very specific use-case - one that should not arise if an ISP is behaving properly (I realise not all do). Perhaps this detail is better included in another how-to somewhere else in the wiki and then linked here, as it presumably has broader application (such as for VLANs).

Greelan (approved):
Amended version looks fine to me. Thanks

fichtner:
Merged, thanks!

PR #445, opened 2022-12-19
added warning about high priority of nat rules

zerwes:
added a **warning** regarding the high prio of nat port forwarding rules and the consequences regarding port forwarding rules without a associated rule.

### DIFF
--- a/source/manual/firewall.rst
+++ b/source/manual/firewall.rst
@@ -122,7 +122,8 @@ Our default deny rule uses this property for example (if no rule applies, drop t
 
 .. Warning::
 
-    `NAT : port forwarding rules <nat.html#port-forwarding>`__ are registered using a priority of 600, i.e. at a very high priority. This has the consequence, that if you use a NAT : port forwarding rule *without a associated rule*, i.e. **Filter rule association** set to **Pass**, no other rules will apply!
+    **NAT rules are always processed before filter rules!**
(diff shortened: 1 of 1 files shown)


AdSchellevis:
I don't think the priority really matters here, nat rules are always processed before filter rules. 

https://github.com/opnsense/core/blob/49f63e8082962c11a43bc58fa8e597149ac1e528/src/etc/inc/filter.inc#L284-L287
https://github.com/opnsense/core/blob/49f63e8082962c11a43bc58fa8e597149ac1e528/src/etc/inc/filter.inc#L379-L386

zerwes:
> nat rules are always processed before filter rules

Yes, this is true, I just stepped into the pitfall with a nat rule and associated filter set to Pass. So I think mentioning the fact that *nat rules are always processed before filter rules* in the place where the priority of the rules is the topic might prevent other from making false assumptions.
This might be clear for some people, but I think a hint at the right place is not hurting.

AdSchellevis:
It certainly doesn't hurt to explain the order of things, I think we just better remove the priority phrase in this case as it doesn't really matter.

zerwes:
OK, got the point @AdSchellevis, sorry, the coin took some time to drop :-/

AdSchellevis:
@zerwes no problem, thanks for the update!

PR #451, opened 2023-02-02
Create apicrud.rst

sde1aney:
basic information around using POST operations for create, read, update, delete operations with the API. worked examples using the Trafficshaper API endpoint, curl syntax.

### DIFF
--- a/source/development/how-tos/apicrud.rst
+++ b/source/development/how-tos/apicrud.rst
@@ -1,0 +1,91 @@
+===============================
+Use the API for CRUD Operations
+===============================
+
+
+Overview
+--------
+
+There are several components which have API capabilities which allow for Create, Read, Update, Delete (CRUD) operations.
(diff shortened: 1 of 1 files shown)


AdSchellevis:
I don't mind extending the examples, but it should fit somewhere in the documentation. The most logic option is likely to add some common scenario's here https://docs.opnsense.org/development/how-tos/api.html

sde1aney:
I'm happy for it to go wherever, I'd put it as another how-to doc to avoid cluttering the existing API how-to page.
I had also considered putting a very small example on the trafficshaper page here [https://docs.opnsense.org/development/api/core/trafficshaper.html](https://docs.opnsense.org/development/api/core/trafficshaper.html), but decided it would then ideally require the equivalent examples on other pages.

AdSchellevis:
a generic example using curl is probably better indeed.

Monviech:
I will close this. If you want to revisit it sometime just ping me and we can reopen it. Thank you.

PR #454, opened 2023-02-13
Update ipv6_dsl.rst

mimugmail:
https://forum.opnsense.org/index.php?topic=32259.msg156936;topicseen#msg156936


fichtner:
I think you mean to add the right thing, but added the wrong setting. Lock interface doesn't do anything. You probably mean "prevent release" setting for interfaces: settings

mimugmail:
Note to self, don't do such things on a mobile :)

fichtner:
Merged, thanks!

PR #459, opened 2023-03-04
Update "first" WireGuard instance number

justinhorton:
The doc references `wg0` as the "first" WireGuard instance; however, this appears to have changed last year, **perhaps inadvertently**, to `wg1`.

I can't find anything referencing this as a problem, but am hoping this doc change PR can serve as a quick confirmation that the change from `wg0` to `wg1` was intended.

---
The following two changes together affected the default instance, so this first changed in 22.7 (assuming I did my git archaeology correctly).

Removed min/max on the plugin's `server->instance` field (22.1):
https://github.com/opnsense/plugins/pull/2760

> [you should be safe, 0 is the default](https://github.com/opnsense/plugins/pull/2760#discussion_r789547565)

Changed `AutoNumberField` default minimum from 0 to 1 (22.7):
https://github.com/opnsense/core/commit/09d782f8f53e169a71b4d800c3325f98de3b2834


AdSchellevis:
due to the new default on autonumber fields (https://github.com/opnsense/core/commit/09d782f8f53e169a71b4d800c3325f98de3b2834), the starting position is 1 now indeed. thanks!

PR #461, opened 2023-03-07
Fixing links only

linuxdude21:
Looks like that last commit my link fixes were not committed.

### DIFF
--- a/source/manual/install.rst
+++ b/source/manual/install.rst
@@ -92,9 +92,9 @@ Depending on your hardware and use case, different installation options are avai
 +--------+---------------------------------------------------+------------+
 
 .. Note::
-   All Full Image types can run both `OPNsense Importer <#OPNsense-Importer>`__
+   All Full Image types can run both `OPNsense Importer <OPNsense-Importer>`__
    before booting into the Live environment and also run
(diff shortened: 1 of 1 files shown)


AdSchellevis:
@linuxdude21 that was more or less intentional, these links don't appear to be functional after compiling, the full links are

linuxdude21:
@AdSchellevis  sorry for my errors.  I think removing "#" should work, which is confusing as I see a lot of info on `<#blabla>`__ working.  Am I dealing with just reStructuredText or does Sphinx play a roll in this?

Is there any way I could compile myself to test before PRing bad syntax?

AdSchellevis:
@linuxdude21 you certainly can, if you plan to contribute more documentation, we should indeed make sure you can validate locally to prevent us needing to compile every cycle.  The https://github.com/opnsense/docs/blob/master/README.md contains the `pip` action to install Sphinx, are you running into issues with that? and if so, which platform are you using

linuxdude21:
@AdSchellevis Great, I will look into this tonight.  Yes I would like to do more documentation for the project.  

Did my last edits not work either?

AdSchellevis:
@linuxdude21 I haven't tried yet, wanted to give you the opportunity to test first locally as well

PR #462, opened 2023-03-08
Update the VMware instruction link to FreeBSD 13.x

tinyboxvk:
The docs should reflect the current FreeBSD version OPNsense is on. There are some differences in the Prerequisites and Installation Steps sections.


AdSchellevis:
@tinyboxvk thanks!

PR #467, opened 2023-03-18
mdns-repeater only supports up to 5 interfaces

da2x:
Upstream documentation:
https://github.com/geekman/mdns-repeater/blob/fbe78e7ffddee8f1dd16a4a990a0b86f94a85964/mdns-repeater.c#L330


fichtner:
Fair enough, thanks!

PR #472, opened 2023-05-18
Added notes dynamicrouting_ospf.rst with alternative configurations and some caveats

rudiservo:
Added notes on disabled networks still adds Area Range config to OSPF.
Added Area Range purpose in a simplistic manner (http://docs.frrouting.org/en/latest/ospfd.html#clicmd-area-0-4294967295-range-A.B.C.D-M-advertise-cost-0-16777215)
Added notes on interfaces for route advertisement with a configured area as an alternative to networks configuration.
Also a small note that interface configuration reduces Firewall Filter Rules has an alternative to networks.

### DIFF
--- a/source/manual/how-tos/dynamicrouting_ospf.rst
+++ b/source/manual/how-tos/dynamicrouting_ospf.rst
@@ -46,6 +46,9 @@ if it is not available. Be careful with this option in HA setups.
 Networks
 --------
 
+.. Note::
+      You can add only the networks that OSPF should send and received routing updates and/or area range summarizations, Networks to be advertized can be configured as Interfaces.
+
 .. image:: images/dynamic_routes_ospf_networks.png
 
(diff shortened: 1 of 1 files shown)


Monviech:
Hello, I have recently reworked the frr docs. If these notes you added are relevant to add to it, please create a new PR and I will review it.

Thanks for your contribution here and sorry it went under the radar.

rudiservo:
Hey, I do have to review if some behavior remains of if it has changed.

i.e. "Disabled networks still add the Area Range for a given Area to the OSPF config file, you can use this for route summarization."

I do think it's a bit important because the configuration behavior of the UI is not very transparent and limiting to FRR OSPF config output or the use of it's full capabilities.
Although I think the OSPF UI needs a rework to make things more 1:1 with the config output.

The rework looks great BTW.

Monviech:
Thank you, if theres some hint to be added in the new docs I'll be glad to review it when offered.

PR #473, opened 2023-05-18
Update logging_firewall.rst: corrected format (bold, italic)

sjjh:
It seems no nested stars are supported, thus concatenated.


AdSchellevis:
@sjjh thanks!

PR #474, opened 2023-05-24
Note about quotes in Sky UK tutorial

boosh:
Add important note to retype quotes in Sky tutorial


fichtner:
It would probably be better to fix the quoting instead of adding this note. Cc @AdSchellevis

boosh:
Yeah ideally. Does RST use backticks for code blocks? It might be better to use those instead of this note.

AdSchellevis:
this https://github.com/opnsense/docs/commit/d1b8f3c680156c9f29506006e9e07f633a2215f3 ?

PR #485, opened 2023-08-23
How-To added for suricata bypass feature 

Monviech:
I added a how-to for the feature https://github.com/opnsense/core/pull/6752 and would like to add it to the docs.


AdSchellevis:
when possible, it would be better to omit screenshots as this eases maintenance in the future.

AdSchellevis:
@Monviech do you mind if I merge your option in the standard documentation around here https://docs.opnsense.org/manual/ips.html#fingerprinting ?

Monviech:
@AdSchellevis I don't mind. Whatever works best for the documentation is fine with me.

AdSchellevis:
@Monviech nevermind, I was looking for a way to glue the feature somewhere, but I think this https://github.com/opnsense/docs/commit/6f042673be4bee208bd65f36f2d7e68e74400df7 should do the trick. Thanks for the document!

PR #488, opened 2023-08-25
Update carp.rst - Add IPv6 CARP Section

Monviech:
I added a section for IPv6 CARP because I couldn't find it in the docs. In the forum there are questions regarding this occasionally.

I was also affected by this in the past because I misinterpreted the section: 
"If you want to add an IPv6 CARP address, write your IPv6 address and the dropdown list will auto-update to 128." 

https://forum.opnsense.org/index.php?topic=33157

I'm open to discussion.

### DIFF
--- a/source/manual/how-tos/carp.rst
+++ b/source/manual/how-tos/carp.rst
@@ -3,7 +3,6 @@
 Configure CARP
 ==============
 
-
 --------
 Overview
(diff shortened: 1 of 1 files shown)


AdSchellevis:
@Monviech thanks, looks good!

PR #492, opened 2023-09-03
Change path for update

xilmen:
Hello, 

A PR to change the path that indicates or upgrade packages. 
The "Check for updates" button is on the status page and not updates page. On Update we only have the Update button. But the check must be check befor (on status page) ?

Have a nice day.


AdSchellevis:
@xilmen nice catch, thanks!

PR #496, opened 2023-09-08
Clarified documentation for IPv6 selective routing

cedricroijakkers:
Slightly elaborated on the use of IPv6 Gateway addresses with respect to [Wireguard selective routing](https://docs.opnsense.org/manual/how-tos/wireguard-selective-routing.html).

The documentation mentions the use of a /127 for IPv6, but is not entirely clear on which address to use for the local tunnel configuration and which one to use for the Gateway.

Added a little note on how to use an IP calculator tool to determine which address is which. Mixing up this configuration will make the VPN not work.


fichtner:
Not sure I get all the trickery employed here. /128 doesn’t work? Usually that is point to point to any gateway. It works for gif/gre devices. Otherwise suggesting a /126 in general might be better.

cedricroijakkers:
Quite honestly, I don't get the trickery either. I was following the guide to set up my routing, and ran into the problem that I was using the wrong IP of the /127 and therefore it didn't work. I have just added some text with the pitfalls I fell into myself, to make sure other people don't.

The rest of the guide is quite accurate, I just specified some details a little more clearly.

Monviech:
Thank you. I have read through it and do not see anything that sticks out as wrong. I merged it.

PR #498, opened 2023-09-10
Wireguard - Reworked wireguard s2s completely & added MTU + MSS to wireguard client

Monviech:
https://forum.opnsense.org/index.php?topic=35877.0

I've seen this a few times so I added a section to the Wireguard site2site and roadwarrior to guide users into configuring MTU and MSS.


FingerlessGlov3s:
@Monviech  why "subtract at least 80 bytes from the Wireguard MTU" should that not be 40 bytes? or is there a reason for 80?

Monviech:
@FingerlessGlov3s 

You are right, I guess I mixed something up. The usual header size for ipv4 is 20 bytes and for ipv6 its 40 bytes. Wireguard header is 32 bytes.
That means with a WAN MTU of 1500, the wireguard MTU is around 1428, and the MSS around 1388.

I just wanted to add some more margin for header extensions but they're not usually used and I dont have any actual numbers for them.

Please give me feedback and I'll adjust the numbers in the PR.

FingerlessGlov3s:
The default WireGuard MTU is 1420, since that'll work for both IPv4 and IPv6 with on a 1500 MTU interface

Then to have a IPv4 TCP connection within the WireGuard tunnel, the highest MSS you can have is 1380, then for IPv6 TCP, it's 1360.

Useful tool, you can keep adding the header sizes for what type of traffic that will be happening.
https://baturin.org/tools/encapcalc/
image

Monviech:
Thats a really useful tool, thanks for sharing. I will double check everything. :+1:

FingerlessGlov3s:
Really you should have two normalization rules one for IPv4 and one for IPv6, so you can squeeze the most out of the tunnel. Plus most people will be doing IPv4 anyways.

I think setting IPv4 or IPv6 on the protocol would do it 🤔, I don't have IPv6 setup on my network at all, to test it.

Monviech:
I mean you could also create different rules for udp and tcp because their header sizes differs too.

My goal was more of a catch all rule in order to mitigate connections not establishing. For performance tweaking you could do a whole bunch of specialized rules, you are right there.

FingerlessGlov3s:
Far as I'm aware you can only MSS clamp on TCP

Monviech:
Yes you are right again. For UDP only the MTU seems to matter.

I was just confused about the fact that you can choose multiple protocols in the normalization rules, but the same mask applies also for dscp etc...

Monviech:
@FingerlessGlov3s
I double checked it and the GUI doesn't allow the IP Protocol being set in the normalization rule.

grafik

FingerlessGlov3s:
There's IPv6 on the list but not entirely sure if that's certain type of traffic or all IPv6. Feel like the IP Version is missing from that page, it should probably be there.

I'd just do one rule saying to use 1360 (PPoE 1352) if you have IPv6 traffic or 1380 (PPoE 1372) if you **only** have IPv4 traffic.

That's probably as good as it gets for now 🙂

source/manual/how-tos/wireguard-s2s.rst
```diff
@@ -8,55 +6,274 @@ WireGuard Site-to-Site Setup
...
+Go to :menuselection:`Firewall --> Rules --> WAN` add a new rule to allow incoming wireguard traffic from Site B.
+
+    ====================== ====================================================================================================
+     **Action**             *Pass*
+     **Interface**          *WAN*
+     **Direction**          *In*
+     **TCP/IP Version**     *IPv4*
+     **Protocol**           *UDP*
+     **Source**             *203.0.113.2*
+     **Source port**        *51820*
```
mimugmail, line 138: Is the source port really 51820?

Monviech: You are right, thanks for the review. It can be any upper port. I will change it to 1024-65535.

mimugmail: I'd rather just remove the line so it's default (-> any)

Monviech: That makes it easier than creating a port alias. I think any is the best choice, too.

Monviech:
@mimugmail 
I have made the requested changes, and I changed a few more details to conform with the new terminology and features.

fichtner:
@Monviech thanks for incorporating the new GUI terminology of 23.7.6 already! came here to say that but now I can shut up :)

AdSchellevis:
@Monviech thanks!

PR #500, opened 2023-09-15
DHCPv6: document downstream prefix delegation

maurice-w:
Documents the existing behaviour as well as the additions made in opnsense/core@a73813684721 and opnsense/core#6853.


fichtner:
Merged, thanks!

PR #507, opened 2023-10-13
Update nat_reflection.rst

Monviech:
- Fixed typo (asynchronous should be asymmetrical traffic)
- Added note about "reply-to" in specific setups with VPN WAN
- Fixed IPsec VTI NAT description


AdSchellevis:
@Monviech thanks!

PR #510, opened 2023-10-31
Update ipsec-swanctl-rw-ikev2-eap-mschapv2.rst

Monviech:
Hint that IPv6 transport doesn't work with UDP Encapsulation

https://github.com/opnsense/core/issues/6964


AdSchellevis:
@Monviech thanks!

PR #512, opened 2023-11-04
Zenarmor CLI install uri update

zenarmor:
Zenarmor CLI insatall script is updated


AdSchellevis:
a bit triggered by the screenshots, but why don't you use the `pkg` install from the command line? (`pkg install os-sunnyvalley` followed by `pkg install os-sensei`)

melroy89:
> a bit triggered by the screenshots, but why don't you use the `pkg` install from the command line? (`pkg install os-sunnyvalley` followed by `pkg install os-sensei`)

If that is the case, why not remove this section from the guide? And replace it with an `pkg install` example?

Monviech:
What is the status with this one?

melroy89:
> a bit triggered by the screenshots, but why don't you use the `pkg` install from the command line? (`pkg install os-sunnyvalley` followed by `pkg install os-sensei`)

Again your documentation page literally explains to use curl instead of `pkg`.. ? https://docs.opnsense.org/vendor/sunnyvalley/zenarmor_cmd_install.html

@AdSchellevis  Any update would be much appreciated. If we shouldn't install it via curl but via the package manager. Please let us know so that we as a community know what we need to change.

AdSchellevis:
@melroy89 best ask @zenarmor, we don't maintain this part of the documentation, but installing zenarmor should be as easy as installing the repo and then checking for updates again.. which I guess is also what [this](https://docs.opnsense.org/vendor/sunnyvalley/zenarmor_install.html) part of the documentation explains.

melroy89:
So you have two ways of installing Zenarmor according to the documentation. One is just using the web interface. And the other documentation page explains how to install zenarmor using the command-line.

This PR is about the command-line documentation page. The OP created this pr, to make the installation up-to-date with the official docs at: https://www.zenarmor.com/docs/installing/installation#installing-on-opnsense

Hence using:

```sh
curl https://updates.zenarmor.com/getzenarmor | sh
```

I therefor believe this PR is correct and this PR should just be approved and merged. Unless @zenarmor of course has any objections?

AdSchellevis:
@melroy89 not sure why you insist on causing more noise than needed, this PR started with a question which wasn't answered (and does deserve one), maybe people from @zenarmor forgot about this one, I don't know. Your reactions are considered rather rude from my end, you might want to consider that when seeking for help in the future.

melroy89:
I don't even use zenarmor. I just found it strange why this pr is open for so long.

mib43:
Not sure why this was left stale for such a long time on our end. We're looking at this.

AdSchellevis:
https://github.com/opnsense/docs/pull/647

PR #517, opened 2023-11-08
Update terminology on Wireguard how-to pages

### LINKED ISSUE #515: Update terminology in Wireguard pages
**Important notices**

- [ *] I have read the contributing guide lines at https://github.com/opnsense/docs/blob/master/CONTRIBUTING.md
- [* ] I have searched the existing issues and I am convinced that mine is new.

**Describe the issue**
Documentation for some of the Wireguard pages in https://docs.opnsense.org/plugins.html#vpn-connectivity have not been updated with the new tab terminology.
https://docs.opnsense.org/manual/how-tos/wireguard-client.html
https://docs.opnsense.org/manual/how-tos/wireguard-selective-routing.html
https://docs.opnsense.org/manual/how-tos/wireguard-client-azire.html
https://docs.opnsense.org/manual/how-tos/wireguard-client-mullvad.html

**Suggestions**

Replace Local references with Instances and Endpoints with Peers and maybe more.

**Version affected**
Documentation version bcb0368

N/A

**Additional context**
N/A

satrapes:
Some time ago the name of some of the tabs in the WireGuard community plugin changed and the changes were not propagated in some pages. This mainly consists of renaming
Local peer -> WireGuard Instance
Endpoints -> Peers
This will close #515

### DIFF
--- a/source/manual/how-tos/wireguard-client-azire.rst
+++ b/source/manual/how-tos/wireguard-client-azire.rst
@@ -30,16 +30,16 @@ favorite text editor.
 Step 2 - Setup WireGuard Instance
 ----------------------------------
 
-Go to tab **Local** and create a new instance. Give it a **Name** and set a desired **Listen Port**. 
+Go to tab **Instances** and create a new instance. Give it a **Name** and set a desired **Listen Port**. 
 If you have more than one server instance be aware that you can use the **Listen Port** only once. In 
 the field **Private Key** insert the value from your text file and leave **Public Key** empty. **DNS** 
-and **Tunnel Address** has also to be taken from the configuration. Hit **Save** and go to **Endpoint** 
+and **Tunnel Address** has also to be taken from the configuration. Hit **Save** and go to **Peers** 
 tab.
 
-On **Endpoint** tab create a new Endpoint, give it a **Name**, set 0.0.0.0/0 in **Allowed IPs** and set
+On **Peers** tab create a new Peer, give it a **Name**, set 0.0.0.0/0 in **Allowed IPs** and set
 the DNS name from your configuration in **Endpoint Address**. Don't forget to do this also for the port.
 
-Go back to tab **Local**, open the instance and choose the newly created endpoint in **Peers**.
+Go back to tab **Instances**, open the instance and choose the newly created peer in **Peers**.
 
 Now we can **Enable** the VPN in tab **General** and continue with the setup.
 
--- a/source/manual/how-tos/wireguard-client-mullvad.rst
+++ b/source/manual/how-tos/wireguard-client-mullvad.rst
@@ -14,7 +14,7 @@ you have read the basic howto :doc:`wireguard-client`.
 Step 1 - Setup WireGuard Instance
 ----------------------------------
 
-Go to tab **Local** and create a new instance. Give it a **Name** and set a desired **Listen Port**. 
+Go to tab **Instances** and create a new instance. Give it a **Name** and set a desired **Listen Port**. 
 If you have more than one server instance be aware that you can use the **Listen Port** only once. In 
 the field **Tunnel Address** insert an unsused private IP address and subnet mask. We don't need it in
 the first step, but as it is required we can't go on without it. Every other field can be left blank.
@@ -23,7 +23,7 @@ Hit **Save** and open your instance again to write down your public key. You nee
 of the configuration from the Mullvad API servers. 
 
 Now change to your OPNsense CLI via SSH or Console and execute *either* of the curl strings below. Please replace
-**YOURACCOUNTNUMBER** with your own ID you got from MullvadVPN and **YOURPUBLICKEY** with the one in your **Local**
+**YOURACCOUNTNUMBER** with your own ID you got from MullvadVPN and **YOURPUBLICKEY** with the one in your **Instances**
 
 The command below is for Mullvad's standard API. DNS requests through a tunnel that uses tunnel IPs generated via this API are "hijacked", so that Mullvad's DNS servers are used to avoid leaks:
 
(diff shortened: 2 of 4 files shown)


source/manual/how-tos/wireguard-client-azire.rst
```diff
@@ -30,16 +30,16 @@ favorite text editor.
...
-and **Tunnel Address** has also to be taken from the configuration. Hit **Save** and go to **Endpoint** 
+and **Tunnel Address** has also to be taken from the configuration. Hit **Save** and go to **Peers** 
 tab.
 
-On **Endpoint** tab create a new Endpoint, give it a **Name**, set 0.0.0.0/0 in **Allowed IPs** and set
+On **Peers** tab create a new Peer, give it a **Name**, set 0.0.0.0/0 in **Allowed IPs** and set
 the DNS name from your configuration in **Endpoint Address**. Don't forget to do this also for the port.
 
-Go back to tab **Local**, open the instance and choose the newly created endpoint in **Peers**.
+Go back to tab **Instances**, open the instance and choose the newly created peer in **Peers**.
```
fichtner, line 42: One note here is that you can also set the instance from the peer with 23.7.8. Might be good to mention this as well or even make it the standard suggestion for shorter workflow?

source/manual/how-tos/wireguard-client-mullvad.rst
```diff
@@ -50,7 +50,7 @@ Also do not forget **Endpoint Address** and **Endpoint Port**. The **Endpoint Po
 
 .. _list: https://www.mullvad.net/en/servers/#wireguard
 
-Go back to tab **Local**, open the instance and choose the newly created endpoint in **Peers**.
+Go back to tab **Instances**, open the instance and choose the newly created peer in **Peers**.
```
fichtner, line 53: Same as previously mentioned.

source/manual/how-tos/wireguard-client.rst
```diff
@@ -8,7 +8,7 @@ Introduction
 
 WireGuard is a simple, fast VPN protocol using modern `cryptography <https://www.wireguard.com/protocol>`__. It aims to be faster and less complex than IPsec whilst also being a considerably more performant alternative to OpenVPN. Initially released for the Linux kernel, it is now cross-platform and widely deployable.
 
-This how-to describes setting up a central WireGuard server on OPNsense and configuring one or more clients to create a tunnel to it. 
+This how-to describes setting up a central WireGuard Instance (server) on OPNsense and configuring one or more client peers to create a tunnel to it.
```
fichtner, line 11:
```suggestion
This how-to describes setting up a central WireGuard Instance on OPNsense and configuring one or more peers to create a tunnel to it. 
```

I'd leave out server/client... it's just confusing if we add more terms that loosely describe this. I think the GUI is better to follow now due to the renaming.

source/manual/how-tos/wireguard-client.rst
```diff
@@ -45,31 +45,31 @@ Step 2 - Configure the local peer (server)
...
-- **Save** the Endpoint configuration, and then click **Save** again
-- Now go back to :menuselection:`VPN --> WireGuard --> Local`
-- Open the Local configuration that was created in Step 1 (eg :code:`HomeWireGuard`)
-- In the Peers dropdown, select the newly created Endpoint (eg :code:`Phone`)
-- **Save** the Local peer configuration again, and then click **Save** once more
+- **Save** the Peer configuration, and then click **Save** again
+- Now go back to :menuselection:`VPN --> WireGuard --> Instances`
+- Open the Instance configuration that was created in Step 1 (eg :code:`HomeWireGuard`)
+- In the Peers dropdown, select the newly created Peer (eg :code:`Phone`)
+- **Save** the Instance configuration again, and then click **Save** once more
```
fichtner, line 72: Same here, one step to select instance from peer and save :)

source/manual/how-tos/wireguard-client.rst
```diff
@@ -18,12 +18,12 @@ Step 1 - Install the WireGuard plugin
 - Once the plugin is installed, refresh the browser page and you will find the WireGuard configuration menu via :menuselection:`VPN --> WireGuard`.
 
 ------------------------------------------
-Step 2 - Configure the local peer (server)
+Step 2 - Configure the Wireguard Instance
```
fichtner, line 21:
```suggestion
Step 2 - Configure the WireGuard instance
```

source/manual/how-tos/wireguard-client.rst
```diff
@@ -18,12 +18,12 @@ Step 1 - Install the WireGuard plugin
...
 ------------------------------------------
 
-- Go to :menuselection:`VPN --> WireGuard --> Local`
-- Click **+** to add a new Local configuration
-- Configure the Local configuration as follows (if an option is not mentioned below, leave it as the default):
+- Go to :menuselection:`VPN --> WireGuard --> Instances`
+- Click **+** to add a new Instance configuration
```
fichtner, line 25:
```suggestion
- Click **+** to add a new instance configuration
```

When not referring to the menu item it's better to use "interfaces" and "peers" consistently in lowercase letters

fichtner (commented):
Thanks for doing this. Most notes are due to existing things that could also use improvement. If you don't want to work on those that's fine. Just let me know and we will pick it up.

source/manual/how-tos/wireguard-client-azire.rst
satrapes: Hmm when is 23.7.8 coming out? I can delay merging this until it comes out so that it matches my personal experience. Don't want to write about something I can't see.

source/manual/how-tos/wireguard-client.rst
satrapes: I am on the fence about this. I think that for newcomers to WireGuard used to other VPN tools terms like client server may make it a bit more intuitive just for the case of the Road Warrior.
I originally added a note that using client and server terms is a bit of a misnomer however it is used intentionally just for the case of the Road Warrior because it may help newcomers understand it a bit better and distinguish between the 2 peers.

fichtner: It's tricky, I agree. We can leave what we have in the docs, but if we start introducing additional terms that have established nomenclature in WireGuard (interface and peer, but fpr us instance and peer to avoid GUI ambiguity) and we also introduce server and client but then fail to use it (because we have better terms given by WireGuard) the question is what the point is.

Maybe it would be better to only mention this once in the main document and leave it out of the how-tos?

satrapes: Basically the nomenclature is established but in the wireguard main page (https://www.wireguard.com/) they also veer from the pure technical terms make use of client and server to allow users to intuitively understand what is going on.
I won't argue too much with you though. can you please update the main page because I am by no means a WireGuard expert and I will update the how-to's by removing mentions of servers and clients.

fichtner:
@satrapes Ok, I'll merge this as is and will modify the pages in a separate pass. Thanks for your work!

PR #518, opened 2023-11-15
Update multiwan.rst

apunkt:
added TIP for upgrading single static IP WAN config to Mutli WAN as it took me hours to find the cause for not working Multi WAN when upgrading from a single static WAN.


AdSchellevis:
@apunkt It's a good idea to add the pre-condition, how about https://github.com/opnsense/docs/commit/0c57a39f65e99757b2fafdd2053bacf76a47e5b5 ?

PR #527, opened 2023-12-28
Unbound: Update predefined blacklist sources

kabili207:
Updated to coincide with my changes in https://github.com/opnsense/core/pull/7103

I also fixed the NoCoin URL to match what was already in place in core


AdSchellevis:
@kabili207 thanks!

PR #533, opened 2024-01-30
Update Spamhaus How-To

thgoebel:
This PR updates the Spamhaus How-To.

1. Update the terms to match the current OPNsense UI (OPNsense 23.7.12). There is no field called "Host(s)" in the "Edit alias" form, it is called "Contents". And in the UI it is called "Refresh frequency", not "update frequency".
2. Mention the DROPv6 list. Copy & pasting the steps for it would ~duplicate~ triplicate the instructions, so I won't do that. I think it is enough to reference it and leave it to the reader to figure out.


AdSchellevis:
@thgoebel thanks!

PR #539, opened 2024-02-19
A minor addition for Hyper-V section in virtuals.rst

cieska:
Hi,
I wanted to suggest a Hyper-V configuration detail when creating the VM. I could not find any better place to add the bit of documentation, as sometimes it can be annoyingly painful when creating a VM on Hyper-V. Let me know if this should be in a different location. 
Cheers.


source/manual/virtuals.rst
```diff
@@ -33,26 +34,27 @@ To install the VMware tools just goto :menuselection:`System --> Firmware --> Pl
...
-------------------
-
+------
 HyperV
 ------
 HyperV Generation 1 and 2 are supported out of the box, no additional drivers
 or tools are needed.
+Secure Boot must not be enabled in the hardware settings for the VM.
+
+.. image:: images/hyper-v_no-enable-sec-boot.png
```
AdSchellevis, line 53: if possible, we would like to avoid screenshots as they tend to change quite often. Describing where one could find a feature often seems to be more resilient to change.

cieska: Should you cancel this request, I will resubmit updated one.

AdSchellevis:
@cieska thanks!

PR #547, opened 2024-03-28
Add instructions in Caddy on how to prevent the HTTP_REFERER error

dmarcoux:
While setting up a reverse proxy for the OPNsense web UI, I stumbled on this error. Without this configuration change, we get the following error after logging in to https://opn.example.com/

> The HTTP_REFERER "https://opn.example.com/" does not match the predefined settings


Monviech:
You are totally right about this. I had that set myself and didnt think about adding it to the guide. Thanks for this. :)

PR #551, opened 2024-04-05
Update models_fieldtypes.rst - Add example to display multiple fields

Monviech:
This example enhances the prior example to display 2 fields instead of 1, using the display_format to format the string.


AdSchellevis:
@Monviech thanks!

PR #554, opened 2024-04-15
Added note to create NAT outbound rule for redirect-gateway to work

cektom:
https://github.com/opnsense/core/issues/7318

### DIFF
--- a/source/manual/how-tos/sslvpn_instance_roadwarrior.rst
+++ b/source/manual/how-tos/sslvpn_instance_roadwarrior.rst
@@ -114,8 +114,6 @@ In order to setup a tunnel on both ends, we need to configure certificates to wa
       it is also possible to clients to share a certificate. When adding a certificate from the user manager the CN is automatically
       set to its name. In this example we will only authenticate using the certificate, no additional user or password will be required.
 
-.. Note::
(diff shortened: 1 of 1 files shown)


source/manual/how-tos/sslvpn_instance_roadwarrior.rst
```diff
@@ -114,6 +114,8 @@ In order to setup a tunnel on both ends, we need to configure certificates to wa
       it is also possible to clients to share a certificate. When adding a certificate from the user manager the CN is automatically
       set to its name. In this example we will only authenticate using the certificate, no additional user or password will be required.
 
+.. Note::
+   In order for the redirect-gateway option to work, a manual NAT outbound rule must be created.
```
AdSchellevis, line 118: Reading through the text, there's no mention of `redirect-gateway` which makes this note a bit weird. How about changing this to a "Tip::" explaining briefly how to send all your traffic via the tunnel, which requires a NAT rule too.

cektom: What do you think about that? 5d08fb412b4b7c972cb826516c21a6ffb618f6b8

AdSchellevis: @cektom looks good, thanks!

PR #563, opened 2024-05-21
Typo fixes in caddy and nat_reflection

Monviech:
A few ":menuselection:" mischaps have been fixed in the caddy and nat_reflection docs. Also a really old one in the nat_reflection doc I maintain but always forgot to fix.


AdSchellevis:
@Monviech thanks!

PR #571, opened 2024-06-09
Create shaper_bufferbloat.rst guide

SeimusS:
Created a guide for FQ_CoDeL to combat Bufferbloat on OPNsense. There are several posts on forum and internet with wrong settings. This should provide proper configuration with optional tuning of FQ_C.


SeimusS:
I tried to make the guide bit more informative as well provide additional information on the parameters. As it looks like a lot of users do something without not knowing why. Or try to look what a parameter does but its hard to find.

Monviech:
I am using such a setup to fight bufferbloat on my asymmetric DSL. (250/40)

What helped me additionally was creating an additional Shaper rule, that matches "tcp (ACK packets only)" and targeting them with an own "Queue TCP ACK Upload" queue with the highest weight (100), which is part of the Upload Pipe that has CoDel enabled.

This seems to reduce latency of TCP handshakes when the Upload is heavily used.

I didn't measure anything though, I've just read at a few places that it would be a good choice to do.

SeimusS:
I've read somewhere the same. However, I myself with the configuration I have, didn't yet seen/experienced latency on Upload during load, so its hard for me to test and write about it. However Download on my side is somewhat tricky (Due to my ISP's technological limitation)

I think the reason (in this specific case) why this Upload or Download latency may happen is due to the **flows** parameter, the default 1024 may be in some use-cases too low, which would cause some specific flows classified into same slots (sub-queues), this would cause for them as well a latency if that specific sub-queue is experience it.

Even if increase of flows parameters could improve it, ultimately I think creating a separate Queue for TCP ACK, which with CoDeL will create sub-queues for specific TCP ACK flows, is ultimately better way. 

_This is just my rough (positive imagination) estimate, but usually the no-knobs deployment for FQ_CoDeL works in 98% of the time if the BW parameter in the Shaper Pipe is set properly. The 2% are then the setups were it needs tuning additional to the out of the box setup of FQ_CoDeL._

I may however maybe try it on Download Pipe to see if there will be improvements. If so we could added it as one of the tuning points into the guide.

I like you Shaper config, very diligent. Thank you for the config.

mimugmail:
The problem is that there is no rule of thumb as it heavily depends on media type (LAN, Ethernet, DSL, Satelliten, Cable) and latency. E.g. for cable with variing upload speeds you may hit congestion earlier than your pipe kicks in

SeimusS:
You are absolutely right, that's the main problem actually.

For the Shaper/Fq_CoDeL to work all the time properly, you need to hit persistently the throughput specified in the BW parameter within the Pipe. Which on these technologies is hardly possible sometimes (or most of the time, really depends). 

Even thou FQ_CoDeL does have mechanisms such as target & interval that measure the delay for each queue and can take action upon it, the fact that you set BW 500 but can reach at a time only 200M will cause a back-pressure which will result in increased latency.

_This specific I tested, and the results was even if this happens the latency is still lower with FQ_CoDel in such situation than without FQ_CoDeL at all. The margin in these situation was 100-140ms (with FQ_CoDel) vs 1200-2000ms (without FQ_CoDeL) during congestion time, worst case scenario._

As for the rule of thumb, there is none, but its recommended to target 85% of advertised ISP BW. Which in theory and most cases should include the worst case scenario throughput. However even this recommendation is recommended to take with grain of salt and needs to be tuned accordingly.

Thus doing few runs of Internet speed-tests, to possibly get average throughput and from it target 85% as the BW value in the Shaper. I would however extent this, to target 85% of the lowest possible throughput as your BW value, if somebody wants to always try prevention of bufferbloat. 

The caveat is, the lowest possible that sometimes occurs maybe waaay to low thus you are loosing substantial amount of BW. Its a heavy tradeoff. For example if I have 500M and lowest throughput occurrence sometimes is 200M. I could set 85% of that 200M as the BW in Pipe. But I am cutting myself of 300M possibly throughput that I am usually able to get just to prevent the occasional latency.

SeimusS:
I did contact the original creators behind FQ_CoDeL and asked them to give potential review for explanation of FQ_CoDeL & its parameters. They gave me some feedback and tips.

I changed this PR to Draft. I will include their feedback into the How-To. As I want to have this as complete and perfect as possible so the wider audience is well informed, and has at least some grasp on what to configure as baseline and tune further if needed.

P.S. Also I found a mistake in the config example, the queue for object pipe shouldn't have any values this is creating a FiFo queue. its not needed as we will separate queue object binded to the specific Pipe.

Regards,
S.

SeimusS:
Few observations/notes from my side (OPNsense configuration):

A. CoDeL can be configured in Pipe
 - If configured in Pipe, standalone Queue object that binds to the Pipe cant be used. Otherwise CoDel will not be applied on the PiPes dynamic queue
 - Enabling the ECN option in Pipe will enable it as well for CoDeL
 - CoDeL is applied to the whole dynamic Queue

B. CoDeL can be configured in Queue
- standalone Queue object bind to the Pipe
- Pipe setting CoDel even if checked has no configuration value when standalone Queue objects are configured
- Enabling the ECN option in Pipe does nothing
- Enabling the ECN option in Queue will enable it as well for CoDeL

C. FQ_CoDel is configured in scheduler
- Can be used with dynamic queues as well non dynamic Queues
- In theory each Flow is in its own Queue, classified by hashing the 5-tuple
- You can enable CoDeL in Queue, ipfw shows its configured (FQ_CodeL in scheduler CoDel in Queue) but doesn't bring any benefit
- Weights in Queues looks like do not have any impact, FQ_CoDeL is not honoring it
- If weighted Queues are needed combination of Codel AQM + (for example) WFQ Scheduler needs to be done
- MASK setting in Queues (same for Pipes) doesn't have any impact, When removed BW equality behaves the same
- Enabling the ECN option in Pipe will enable it as well for FQ_CoDeL in scheduler
- Enabling the ECN option in Queue does nothing (if CoDeL is disabled)

SeimusS:
Hello @AdSchellevis 

I think I am finished here.  I was lucky enough to reach towards FQ_CoDel creators & bufferbloat community. They were kind enough to review fully the docs adjust it and provide more proper details about FQ_CoDel.

Special thanks to @richb-hanover & @dtaht for giving me their time, advice and full review!

Regards,
S.

AdSchellevis:
@SeimusS very nice writeup, thanks for contributing it. I'll merge and push it to the docs.

richb-hanover:
PS @SeimusS  - you're famous(-ish)! https://randomneuronsfiring.com/opnsense-fights-bufferbloat/

Thanks again for letting me contribute.

SeimusS:
@richb-hanover hahaha unexpected & lovely of you. Many thanks for mentioning OPNsense (& its official docs for FQ_C) on your blog!

I had fun too.

Regards,
S.

SeimusS:
@AdSchellevis 

One question I didn't put the reference for the doc "shaper_bufferbloat" into source/manual/shaping.rst its only in the source/manual/how-tos/shaper.rst.

Which makes it maybe a bit hidden when somebody looks at the Traffic Shaping manual as whole. Is it possible you will additionally put it there as well? Otherwise I can open another PR.

`
how-tos/shaper_bufferbloat`

Regards,
S.

AdSchellevis:
@SeimusS I missed that as well, just added https://github.com/opnsense/docs/commit/39640f28abcc34892a14c8e034951f773c3ed288 which should help.

richb-hanover:
@AdSchellevis @SeimusS [Google Alerts for "bufferbloat"](https://www.google.com/alerts) points me to this video: 

https://www.youtube.com/watch?v=x05Nxw4E8fI

Enjoy!

PR #572, opened 2024-06-11
security/strongswan: Update docs about udp transport for roadwarrior setups

Monviech:
Adjust note about udp transport with ipv6
https://github.com/opnsense/core/issues/6964#issuecomment-2160192781


fichtner:
Let's keep this parked for now. Don't want to hype something that's in 24.7 but may not because FreeBSD keeps it away from FreeBSD 14. The wording on this should be simpler eventually but let's revisit in the 24.7 release week when we know more. CC @AdSchellevis

PR #573, opened 2024-07-03
Update sfp_compatibility.rst and ice driver tunable

paolba:
only worked after enabling the ice modules with the tunables. maybe add an information about this to the documentation. I only found this info in the forums. without the tunables i was only able to get a 10Gbase connection when setting the speed on the switch.

tunables:
if_ice_load="YES"
ice_ddp_load="YES"

reboot after setting the tunables. nothing more is needed.

source:
https://forum.opnsense.org/index.php?topic=30759.0

### DIFF
--- a/source/hardware/sfp_compatibility.rst
+++ b/source/hardware/sfp_compatibility.rst
@@ -205,4 +205,5 @@ Uptimed   UP-SFP28-SR-CI                 25G
 Vendor    Type                           Speed   Notes
 ========= ============================== ======= =========================
 FlexOptix P.C3025G.H Passive             25G
+FS        SFP-H25G-CU1M                  25G     With Intel compatibility
 ========= ============================== ======= =========================


AdSchellevis:
you can add `ice_ddp_load` to the tunables, `ice_load` shouldn't be needed  if I'm not mistaken as the module is loaded by default (our default factory default config files only include `ice_ddp_load`)

paolba:
> you can add `ice_ddp_load` to the tunables, `ice_load` shouldn't be needed if I'm not mistaken as the module is loaded by default (our default factory default config files only include `ice_ddp_load`)

thanks for the hint! I can confirm 25Gbase still working after removing the tunable "if_ice_load". Do you know why "ice_ddp_load" is necessary and why this is not enabled by default?

AdSchellevis:
it's an intel driver thing, firmware is only loaded when requested (the driver works without it, but **very** slow)

PR #575, opened 2024-07-09
Only having Hardware acceleration setting option

melroy89:
- Remove other cryptography settings, since these seem be missing/ not present anymore in the WebUI (correct me if I'm wrong!!). I'm using OPNsense v24.1.
- I only see "Hardware acceleration" as an option, so I only keep this setting.

image


AdSchellevis:
@melroy89 thanks, looks like we missed a spot there indeed

PR #577, opened 2024-07-10
Small grammar & layout fixes on BE page

melroy89:
See title.

### DIFF
--- a/source/be.rst
+++ b/source/be.rst
@@ -9,7 +9,7 @@ Business Edition
 
 
 
-A mission critical version of the well known OPNsense firewall.
+A mission critical version of the well-known OPNsense firewall.
 
 The Business Edition offers additional safeguards where functional changes are being included in a
 more conservative manner and feedback has been collected from development and community.
--- a/source/be.rst
+++ b/source/be.rst
@@ -9,24 +9,24 @@ Business Edition
 
 
 
-A mission critical version of the well known OPNsense firewall
+A mission critical version of the well known OPNsense firewall.
 
 The Business Edition offers additional safeguards where functional changes are being included in a
 more conservative manner and feedback has been collected from development and community.
 
 Offering specific business-oriented features and third party security verification.
-Currently, the only open source LINCE compliant firewall.
+Currently, the only open-source LINCE compliant firewall.
 
 * Mission critical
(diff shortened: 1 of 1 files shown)


source/be.rst
```diff
@@ -9,24 +9,24 @@ Business Edition
...
 
-A mission critical version of the well known OPNsense firewall
+A mission critical version of the well-known OPNsense firewall.
 
 The Business Edition offers additional safeguards where functional changes are being included in a
 more conservative manner and feedback has been collected from development and community.
 
 Offering specific business-oriented features and third party security verification.
-Currently, the only open source LINCE compliant firewall.
+Currently, the only open-source LINCE compliant firewall.
```
AdSchellevis, line 18: I don't  think a hyphen is needed for open source (https://en.wikipedia.org/wiki/Open_source)

source/be.rst
```diff
@@ -9,24 +9,24 @@ Business Edition
...
 
 * Mission critical
 * LINCE compliant (security verification by trained third party independent professionals)
 * Commercial firmware repository
 * Free GeoIP database
-* Official OPNsense Open Virtualisation Image
+* Official OPNsense Open Virtualization Image
```
AdSchellevis, line 24: we prefer British english, in which case an `s` is used

AdSchellevis (changes requested):

source/be.rst
melroy89: ah that is fine. OK

source/be.rst
melroy89: https://learn.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/o/open-source

melroy89: Since it's a noun, I do believe you could/can use hyphenate the word open source as in open-source software or in this case open-source LINCE compliant firewall .

AdSchellevis: I don't think either spelling is wrong, but if both are correct, it's a matter of preference in which case we can keep it like it is. (I don't have a strong preference, but also like to prevent the next guy opening a PR in a year or two to change it back)

melroy89:
@AdSchellevis You should be able to merge it now.

PR #582, opened 2024-07-31
Update wazuh-agent.rst

mimugmail:
Small note to adjust agent id and use a rule which already exist and works


AdSchellevis:
@mimugmail thanks, out of curiosity, are you using the plugin regularly?

mimugmail:
No, only syslog. But had a customer with demand to use active response. With this it will work out of the box  (bad for Consulting business) :)

AdSchellevis:
> ... (bad for Consulting business) :)

oops, sorry 🫣

PR #593, opened 2024-08-19
Update helloworld.rst

TotalGriffLock:
Correct a small coding error in the example SettingsController jQuery code. When performing form validation the code tries to retrieve a collection called "general" but general is the name of the form, the collection should start "helloworld".

The correct code is present in the example github repo, but incorrect in the docs.


AdSchellevis:
@TotalGriffLock thanks, we should probably refactor the hello world example at some point, most of this stuff is wrapped inside `ApiMutableModelControllerBase` nowadays, which would shorten the text a lot and helps people on focusing on the things most commonly used.

TotalGriffLock:
That would be good, I've written 2 plugins based on the hello world example so you'd be saving me some time!

Issue #592, opened 2024-08-19
Add Boot Environment [Snapshots] documentation


Monviech:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/docs/blob/master/CONTRIBUTING.md
- [x] I have searched the existing issues and I am convinced that mine is new.

**Is your feature request related to a problem? Please describe.**

Documentation for: https://github.com/opnsense/core/pull/7749

**Describe the solution you like**

- Short explanation about what boot Environments are
- How to create a Snapshot
- How to restore a Snapshot
- Usecases (like creating a Snapshot before doing a Major Upgrade)

**Describe alternatives you considered**

N/A

**Additional context**

N/A

PR #597, opened 2024-09-02
system/snapshots: Add initial documentation for boot environments

Monviech:
Fixes: https://github.com/opnsense/docs/issues/592

### DIFF
--- a/source/manual/snapshots.rst
+++ b/source/manual/snapshots.rst
@@ -4,9 +4,11 @@ Snapshots
 
 .. contents:: Index
 
+
 --------------
 Quick Overview
 --------------
+
 *Snapshots* capture the base system at a given moment. They're useful for administration, as they introduce a fallback boot option when more risky operations like major upgrades or experimental configurations fail. If something goes wrong, it is quite simple to roll back to the previous working state.
 
 These *Snapshots* leverage ZFS features, making them space-efficient and easy to manage. During the boot process, the latest and previous saved *Snapshots* can be chosen with *Option 8*, effectively switching between different system states.
--- a/source/manual/snapshots.rst
+++ b/source/manual/snapshots.rst
@@ -31,7 +31,7 @@ Recommended Workflow
 * OPNsense works as expected: 
     * Remove the **known-good** *Snapshot* and keep using the **default** *Snapshot*
 * OPNsense has issues: 
-    * Activate the **known-good** *Snapshot*  either in the WebGUI or the Boot Menu
+    * Activate the **known-good** *Snapshot* either in the WebGUI or the Boot Menu
     * After rebooting, remove the bad **default** *Snapshot*
     * Rename the **known-good** *Snapshot* back to **default**
 
--- a/source/manual/snapshots.rst
+++ b/source/manual/snapshots.rst
@@ -15,7 +15,7 @@ These *Snapshots* leverage ZFS features, making them space-efficient and easy to
 
 The *Snapshots* are differential. A new *Snapshot* is always the latest current running state. This means, there are no incremental *Snapshot* chains, as seen in other implementations such as Hypervisors.
 
-.. Note:: Each Snapshot can be created and deleted as its own self contained entity with no other dependencies. For practical reasons, it is recommended to keep no more than two *Snapshots*. Snapshots can use a considerable amount of space if they are not cleaned up after use.
+.. Note:: Each *Snapshot* can be created and deleted as its own self contained entity with no other dependencies. The size of each *Snapshot* will grow over time as the distance in bytes between the current *active Snapshot* and all stored *non-active Snapshots* grow.
 
 .. _snapshot-recommended-workflow:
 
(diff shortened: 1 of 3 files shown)


Monviech:
@SeimusS Can you give this a look? I tried to stay pragmatic in these docs, giving people what I think they need when they require Snapshots.

fichtner:
The idea was to ping @sheridans but more eyes are better ;)

Monviech:
@fichtner Oops but the more the merrier. :)

SeimusS:
Well you looped me so here is my feedback :)


**Quick Overview**

> Boot Environments require ZFS as filesystem. UFS will not work since it does not support file system level snapshots.

I would change the first "NOTE" to "Important" in the quick overview so it pokes into peoples eyes

**Recommended Workflow**
Is it necessary, to activate the new Snapshot? isn't better to continue on the current active = default and in case of problems to rollback on the created one?

Also depending on the issues that can be experienced rollback can be as well GUI, I think explixitly pointing to do it via CLI in boot menu may create a bit stress for some users (believe me or not I know few people that are scared to type cat command in the CLI)

**Creating a Snapshot**
I think this sentence is bit confusing >

> Now, there should be one Snapshot named default which is Active N, and a new Snapshot with a timestamp that is Active R.

Because >
N means "active now"
R means "active on reboot"

Regards,
S.

Monviech:
@SeimusS Thanks for the feedback.

- Changing the first note to attention would be a good choice
- For the workflow, I have used this as source, it seems to make sense: https://forums.FreeBSD.org/threads/bectl-confusion.85295/post-568756
- I'll make the sentence about the "Active" more clear
- My target audience were system administrators who do an upgrade of a hardware appliance. They should not be scared of the Boot Menu (I think)

source/manual/snapshots.rst
```diff
@@ -0,0 +1,107 @@
...
+--------------
+*Snapshots* capture the base system at a given moment. They're useful for administration, as they introduce a fallback boot option when more risky operations like major upgrades or experimental configurations fail. If something goes wrong, it is quite simple to roll back to the previous working state.
+
+These *Snapshots* leverage ZFS features, making them space-efficient and easy to manage. During the boot process, the latest and previous saved *Snapshots* can be chosen with *Option 8*, effectively switching between different system states.
+
+.. Attention:: **Snapshots require ZFS as filesystem**. UFS will not work since it does not support file system level snapshots.
+
+The *Snapshots* are differential. A new *Snapshot* is always the latest current running state. This means, there are no incremental *Snapshot* chains, as seen in other implementations such as Hypervisors.
+
+.. Note:: Each Snapshot can be created and deleted as its own self contained entity with no other dependencies. For practical reasons, it is recommended to keep no more than two *Snapshots*. Snapshots can use a considerable amount of space if they are not cleaned up after use.
```
AdSchellevis, line 18: Why two? If I'm not mistaken the storage amount used is similar if snapshots "stack", so, when it looks like this in time:

```
[1] ---------------------- [2] 
```

It would roughly require the same as 

```
[1] ----------[2]---------- [3] 
```

As the distance between 1 and 3 is the same as the distance (in bytes) between 1 and 2 of the earlier example. This obviously changes when [1] is forked multiple times, like:

```
   [1] ------ [2] ------- [3]
             \
              ------- [4] 
                             \
                              ---- [5]
```

source/manual/snapshots.rst
```diff
@@ -0,0 +1,107 @@
...
+#. Create a new Snapshot
+#. Activate the new Snapshot
+#. Do administrative changes (e.g. firmware upgrade)
+#. Reboot and test the changes:
+
+* OPNsense works as expected: 
+    * Remove the old *default* Snapshot
+    * Rename the current active Snapshot to *default*
+* OPNsense has issues: 
+    * Select the old *default* Snapshot in the Boot Menu
```
AdSchellevis, line 36: if you still are able to boot, you might want to prevent using the boot menu and just press "active" in the ui (and reboot).  In most cases you're likely still able to use the gui, it's just not what you expected.

AdSchellevis (changes requested):

SeimusS:
>     * Changing the first note to attention would be a good choice
This is just eye candy but from my experience of writing documentation it makes Huge difference

>     * For the workflow, I have used this as source, it seems to make sense: https://forums.FreeBSD.org/threads/bectl-confusion.85295/post-568756
It may and it does, but it doesn't mean its optimal. The less steps is required to do something the less is the probability of a user mistake. Simplicity is what should be targeted if possible or to try to crate simplicity. From What I read here for me it looks like there are just unnecessary steps than can create in some scenario for some people problems thus mistakes.

I overall think the snap should be create as rollback, and only be activated in case of problems.

This is the beauty of writing documentation, you can by your own experience create more optimal version.

>     * I'll make the sentence about the "Active" more clear
Thanks!
 
>     * My target audience were system administrators who do an upgrade of a hardware appliance. They should not be scared of the Boot Menu (I think)
Thats true, but keep in mind the OPNsense audience is more than just system or network admins. There is a lot of beginners newcomers that will use this feature ;)

@Monviech 

Regards,
S.

Monviech:
@SeimusS I have rewritten most of it to improve the logic and make it a bit more beginner friendly. Maybe its better now than before?

SeimusS:
@Monviech 

I was just looking at it, now looks much more beginner friendly and simpler (and hopefully less prompt to beginners mistake).
I like it!

One question tho, in the docs is mentioned snap will rollback
A. OS version in case upgrade was happened
B. Configuration changes/differences

Will the rollback of snap, rollback as well any installed packages, custom scripts (like those in rc. & syshook) etc? Because if yes maybe it would be good explicitly point what is part of the rollback and what is not. Cause this may sounds stupid, but I can imagine a situation where an user may ask these questions on forum (best case scenario) or assume it will/will not and than be surprised (worst case scenario) and open a topic on the forum. Could cause a bit of Catch-22 scenario.

Regards,
S.

Monviech:
Hmm...

@fichtner is the kernel a part of a boot environment?

If an upgrade includes a kernel update, would the old kernel need to be manually selected in the Boot Menu in addition to the Boot Environment?

fichtner:
The kernel is also stored on the ZFS BE. How the boot loader handles this half-way switching BEs and also switching kernels I don't know but it doesn't matter for the GUI use case we want to document.

Monviech:
Let's go for it, if anybody has improvements, PRs are welcome.

Merge discussed with @AdSchellevis

Issue #590, opened 2024-08-17
Interfaces / Other types / VLAN


AdSchellevis:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/docs/blob/master/CONTRIBUTING.md
- [x] I have searched the existing issues and I am convinced that mine is new.

**Is your feature request related to a problem? Please describe.**

Not a solid plan yet, but it might be practical to explain a bit about vlan usage and best practices like not combining tagged and untagged traffic on the same interface (as this will clutter stats and has some downsides in terms of visibility)

**Describe the solution you like**

Maybe extend the VLAN part a bit, but adding some general documentation about interfaces and common types might also be practical.

**Describe alternatives you considered**

none

SeimusS:
It doesn't only cluster stats in many cases it can be a cause of interface errors. Have seen some examples on the forum already.

I think this is a great idea btw. A bit of more educative approach targeted for newcomers maybe? 

How deep you want to go into the VLAN rabbit hole?
Examples and best practices only or as well explanation on a more technical level? ;)

Regards,
S.

PR #610, opened 2024-10-08
interfaces: vlan/lagg - Add tutorial section

Monviech:
I wrote this from the perspective of connecting an OPNsense Appliance to a Managed Switch, one of the most common scenarios for network administrators.

It explains how such a setup should be tackled, preventing common mistakes like mixing tagged/untagged VLANs or creating a network bridge instead of using LAGG when more than one physical interface should be connected to the same switch.

Recommendations for naming schemes for easier administration are included.

Fixes: https://github.com/opnsense/docs/issues/590


source/manual/how-tos/vlan_and_lagg.rst
```diff
@@ -0,0 +1,191 @@
...
+- | **Untagged VLAN:** Frames are sent without VLAN tags. The switch port assigns incoming untagged frames to a default VLAN.
+    Untagged VLANs are typically used on **access ports** connected to end devices that are not VLAN-aware, such as computers.
+
+
+Each VLAN will represent its own isolated network, connected by a VLAN-aware router like the OPNsense.
+Traffic that should cross VLAN boundaries must be routed and controlled via firewall rules. This is known as `Inter-VLAN-Routing`.
+
+.. Note::
+
+    Not all switches support VLAN. `Unmanaged Switches` only provide basic functionality.
```
AdSchellevis, line 34:
```suggestion
    Not all switches support VLANs. `Unmanaged Switches` only provide basic functionality.
```

source/manual/how-tos/vlan_and_lagg.rst
```diff
@@ -0,0 +1,191 @@
...
+
+
+1. Setup LAGG interface
+-------------------------
+
+See the section on `LAGG </manual/other-interfaces.html#lagg>`_ for more details.
+
+.. Note::
+
+    This step is optional but highly recommended. It will create an abstraction layer between the VLAN and the physical interfaces,
```
AdSchellevis, line 84: I'm not sure about the highly recommended part, I think it depends per deployment. For example, when installing a small branch office, you might want to isolate the guest wifi network, but rather prevent more complexity when only connecting with a single cable to the switch in the foreseeable future.  Sometimes it's just better starting small, as often it requires more knowledge as well.  Just my tow cents on the subject.

AdSchellevis (changes requested):

AdSchellevis:
@Monviech looks good, one small typo and a remark. Thanks!

source/manual/how-tos/vlan_and_lagg.rst
Monviech: Sure that makes sense, I will change the recommendation accordingly.

SeimusS:
@Monviech 

Bit late to the party as I was out for quiet a bit. But few small comments:

**Fast timeout** 
- this one can be actually tricky, and I would not opt it as recommended. Because depending on Vendors implementation it can cause a lot of troubles and break the LACP. Famous example is Juniper LACP fast mode. Where JP misses the LACP hello intervals effectively breaking the heartbeat causing the LACP and LAGG to go down. Just something to keep in mind.

**Hash Layers**
- do not need to be the same between two peering devices, it can be considered as a best practice but its not a rule of must be.

**If multiple switches are in a Stack configuration, using aggregation protocols like MLAG (Multi-chassis Link Aggregation) can be required on the switch side.**
- MLAG or MEC, is not the same as Switch stack. A Switch stack usually doesn't support MLAG or MEC. Stack of switches are basically several independed switches that are managed as one, but their data and control plane is handled individually. 

Overall very nice guideline and I am happy to see you use the term "Inter-VLAN-Routing"!

Regards,
S.

Monviech:
@SeimusS Thanks for looking over it. Wanne do a PR where you add these suggestions?

SeimusS:
@Monviech 
Sure, if its fine with you, I will open a PR and do some adjustment

PR #618, opened 2024-10-23
Fix support for newer sphinx_rtd_theme versions up to 2.0

tinyboxvk:
Took the idea from https://github.com/readthedocs/sphinx_rtd_theme/issues/1240 and resolved the issue in https://github.com/opnsense/docs/commit/15be11c750ae9738d6e6427b8cf92e5dd0103ba0.


AdSchellevis:
@tinyboxvk nice catch! thanks

PR #619, opened 2024-10-23
Update README.md - Update build process on FreeBSD 14

tinyboxvk:


AdSchellevis:
@tinyboxvk I haven't tried these instructions on my end, but assume you did, the changes look reasonable.

tinyboxvk:
Yes, I have. It was on a new FreeBSD 14.1 install with minimal tools pre-installed.

README.md
```diff
@@ -57,17 +57,17 @@ executed using:
 #### Generate HTML documents
 
 ```
-make html
+gmake html
```
fichtner, line 60: I wonder when this broke but it's correct indeed.

README.md
```diff
@@ -32,16 +32,16 @@ that it owns or licenses any of the mentioned, nor does it grant them.
 
 #### Prepare build
 
-On FreeBSD the following packages are required:
+On FreeBSD 14 the following packages are required:
```
fichtner, line 35:
```suggestion
On FreeBSD 14 the following packages are required:
```

Personally I would leave this as is and just fix the steps. They will never be perfect.

README.md
```diff
@@ -32,16 +32,16 @@ that it owns or licenses any of the mentioned, nor does it grant them.
...
 
 ```
-pkg install py39-pip jpeg-turbo gmake
+pkg install python py311-pip jpeg-turbo gmake
 ```
 
 Install Sphinx, our default theme and contrib packages:
 
 ```
-pip[3] install -r requirements.txt --upgrade
```
fichtner, line 44: I'm not sure if the [3] came from FreeBSD or Linux. On FreeBSD it's "pip" indeed, but this step is supposed to be run on FreeBSD or Linux

README.md
tinyboxvk: Thanks for the review. I have added a commit to this PR to remove the exact version number.

README.md
tinyboxvk: I just tested on Ubuntu 24.04. The `python3-pip` package installs both `pip` and `pip3` to `/usr/bin`. They are the exact same file with the same hash.

fichtner: ok great, thanks!

README.md
```diff
@@ -35,13 +35,13 @@ that it owns or licenses any of the mentioned, nor does it grant them.
 On FreeBSD the following packages are required:
 
 ```
-pkg install py39-pip jpeg-turbo gmake
+pkg install python py311-pip jpeg-turbo gmake
```
fichtner, line 38: one last note here is that python may conflict with py311- as one is implicit but all other packages follow the version/ py311-pip should install the right python311 package and avoid the metapackage?

tinyboxvk: There is a difference in what packages are installed.

```
root@freebsd:~ # pkg install python311
...
New packages to be INSTALLED:
        libffi: 3.4.6
        mpdecimal: 4.0.0
        python311: 3.11.10
        readline: 8.2.13_1

Number of packages to be installed: 4
...
```

This does not create `python` or `python3` under `/usr/local/bin`, so scripts like `collect_api_endpoints.py` wouldn't run.

```
root@freebsd:~ # pkg install python
...
New packages to be INSTALLED:
        libffi: 3.4.6
        mpdecimal: 4.0.0
        python: 3.11_3,2
        python3: 3_4
        python311: 3.11.10
        readline: 8.2.13_1

Number of packages to be installed: 6
...
```

This creates `python` and `python3` under `/usr/local/bin`, where `python` is a symbolic link to `python3`, and `python3` a symbolic link to `python3.11`.

tinyboxvk: And just to add, `collect_api_endpoints.py` for example depends on `/usr/local/bin/python3` being available, which is installed by the metapackage.

AdSchellevis:
@tinyboxvk thanks

PR #621, opened 2024-10-26
Fix typos in ipsec-swanctl-rw-ikev2-eap-mschapv2.rst

aaronyeung:
Fixes typos in `source/manual/how-tos/ipsec-swanctl-rw-ikev2-eap-mschapv2.rst`

A typo in length, and case of OPNsense.


AdSchellevis:
@aaronyeung thanks!

PR #624, opened 2024-10-27
Select OpenVPN CA as issuer of OpenVPN user cert

ilumos:
Also make steps use same formatting as above two paragraphs and fix minor formatting issues


AdSchellevis:
@ilumos thanks!

PR #626, opened 2024-10-30
Minor changes and updates

Bishop-trevorstuart:
Changed a typo regarding adding DNS to a peer. 
Also added additional direction to a note regarding gateway configuration.

### DIFF
--- a/source/manual/how-tos/wireguard-client-mullvad.rst
+++ b/source/manual/how-tos/wireguard-client-mullvad.rst
@@ -40,8 +40,7 @@ The alternative command below is for Mullvad's other API. DNS requests through t
 What you receive is what WireGuard calls **Allowed IP** for your WireGuard Instance. Edit your instance again and remove
 the value of **Tunnel Address** you used when setting it up and change it to the one you got.
 
-On **Peers** tab create a new Peer, give it a **Name**, set 0.0.0.0/0 in **Allowed IPs** and set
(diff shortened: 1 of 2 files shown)


source/manual/how-tos/wireguard-client-mullvad.rst
```diff
@@ -40,8 +40,7 @@ The alternative command below is for Mullvad's other API. DNS requests through t
 What you receive is what WireGuard calls **Allowed IP** for your WireGuard Instance. Edit your instance again and remove
 the value of **Tunnel Address** you used when setting it up and change it to the one you got.
 
-On **Peers** tab create a new Peer, give it a **Name**, set 0.0.0.0/0 in **Allowed IPs** and set
-the **DNS** to 193.138.218.74. This is the one MulladVPN provides for privacy.
```
AdSchellevis, line 44: Why is the dns removed here? I'm not a MulladVPN user, but this seems to be part of the service.

Bishop-trevorstuart: you don't add DNS in the peers section of wireguard at all. So it's misleading to put it here. 
Also from what I've read the DNS server itself is actually changed now too, so it's not just in the wrong place, it's also incorrect info...

AdSchellevis: You're right, not sure what the intention was here then, let's remove it.

PR #631, opened 2024-11-06
Tayga: specify interface for outbound NAT rule

maurice-w:
Apparently, the default setting isn't always "WAN":
https://forum.opnsense.org/index.php?topic=43685.msg218520#msg218520


fichtner:
Merged, thanks!

PR #633, opened 2024-11-08
MaxMind - added in accountID update

calvolson:
- Added instructions to update account ID as it is needed for the URL to work. 
- Updated the reference on the license key to exactly match the example URL.
- Cleaned up the sentence.


AdSchellevis:
@calvolson thanks!

PR #634, opened 2024-11-09
manual: user must 'Apply' changes after generating wg peer

omegadrh:
After clicking "Store and generate next," `wg showconf wg1` on opnsense doesn't show the new peer (and connection attempts from the peer fail without explanation). Visiting the WG->Peers page and clicking "Apply" fixes this.

I think it'd be better for "Store and generate next" to automatically apply the new peer to the wg config, but until that's the case, at least the docs can describe a procedure to make it work.


AdSchellevis:
@omegadrh thanks!

PR #635, opened 2024-11-12
Document Radius group sync

adn77:
This tries to better explain how the *Synchronize groups* feature works with RADIUS servers.

https://github.com/opnsense/core/issues/6230#issuecomment-2463987192


source/manual/how-tos/user-radius.rst
```diff
@@ -15,13 +15,26 @@ Fill in the form:
 **Services offered**            Authentication *Select Authentication,for Captive portal + accounting*
 **Authentication port value**   1812           *Port number, 1812 is default; for accounting it's 1813*
 **Authentication Timeout**      5              *Timeout for Radius to respond on requests*
-**Synchronize groups**                         *Enable to Synchronize groups, requires the option above*
-**Limit groups**                               *Select list of groups that may be considered during sync**
-**Automatic user creation**                    *When groups are automatically synchronized,
-                                               this offers the ability to automatically create the
-                                               user when it doesn't exist.*
+**Synchronize groups**                         *Enable to read groups from RADIUS server - requires the
+                                               CLASS attribute to return the designated (single!) group**
```
AdSchellevis, line 19: single is not necessarily true, when multiple classes are returned, each starting with `CN=`, all will be passed.

https://github.com/opnsense/core/blob/c7f9b4b13ac172d94a9053b757ab42e7c4ff76f7/src/opnsense/mvc/app/library/OPNsense/Auth/Radius.php#L524

https://github.com/opnsense/core/blob/c7f9b4b13ac172d94a9053b757ab42e7c4ff76f7/src/opnsense/mvc/app/library/OPNsense/Auth/Base.php#L166

source/manual/how-tos/user-radius.rst
```diff
@@ -15,13 +15,26 @@ Fill in the form:
...
+**Limit groups**                               *Limit user authentication to members of these groups
+                                               (user has to pre-exist in OPNsense for this to work,
```
AdSchellevis, line 21: I don't think that's true, below the authentication steps :

1) call authenticate:
https://github.com/opnsense/core/blob/c7f9b4b13ac172d94a9053b757ab42e7c4ff76f7/src/opnsense/mvc/app/library/OPNsense/Auth/Radius.php#L411

2) push found groups to `setGroupMembership()`  including flag to create user

https://github.com/opnsense/core/blob/c7f9b4b13ac172d94a9053b757ab42e7c4ff76f7/src/opnsense/mvc/app/library/OPNsense/Auth/Radius.php#L535

3) local user group (doesn't exist) will not match requested set of groups

https://github.com/opnsense/core/blob/c7f9b4b13ac172d94a9053b757ab42e7c4ff76f7/src/opnsense/mvc/app/library/OPNsense/Auth/Base.php#L183-L190


In a future release we probably will move creation a little bit more on top so it will sync when no groups are requested, but if the radius or ldap server sends a list, it will create the user now.

source/manual/how-tos/user-radius.rst
```diff
@@ -15,13 +15,26 @@ Fill in the form:
...
+                                               to be enabled and actually return a group for a user.*
 ============================== =============== =========================================================
 
+.. Note::
+   RADIUS does not support a *memberOf* group concept by design. OPNsense uses the returned CLASS attribute
+   instead to find a string containing the user's group membership. Since the **Synchronize groups**
+   feature shares the same code of the LDAP server feature **Synchronize groups** the string defined as
+   CLASS value must be prefixed with *CN=* (e.g. *CLASS="CN=MyVPN-Group"*)!
+
+   Additionally the group separator must be a line break (*\n*). As most RADIUS servers will not support
```
AdSchellevis, line 35: If I'm not mistaken, you can send multiple the same attributes using radius, in which case they will be merged accordingly. I'm pretty sure I tested this years ago, currently I don't have a setup on my end.

AdSchellevis (changes requested):

source/manual/how-tos/user-radius.rst
adn77: The separator is a line break. Some Radius servers will not allow special characters in the Class string.
That's what https://github.com/opnsense/core/issues/6230 was about.

I added the asterisk to explain this case below...

source/manual/how-tos/user-radius.rst
adn77: I only have MS NPS to test with. At least there it's not possible.
Do you want me to switch *most* with *some*?

AdSchellevis: Maybe better rephrase, the newline character is likely not what the radius server sends in this case.  I'm certainly fine with mentioning most Radius servers will only support sending a single entry, I'm not sure which ones are able to send more and it has been ages since I used this.

source/manual/how-tos/user-radius.rst
adn77: Oh, I'll change that.
I actually was forcing the group in OpenVPN. This failed when the user didn't already exist.

source/manual/how-tos/user-radius.rst
```diff
@@ -15,13 +15,24 @@ Fill in the form:
...
+                                               to be enabled and actually return a group for a user.*
 ============================== =============== =========================================================
 
+.. Note::
+   RADIUS does not support a *memberOf* group concept by design. OPNsense uses the returned CLASS attribute
+   instead to find a string containing the user's group membership. Since the **Synchronize groups**
+   feature shares the same code of the LDAP server feature **Synchronize groups** the string defined as
+   CLASS value must be prefixed with *CN=* (e.g. *CLASS="CN=MyVPN-Group"*)!
+
+   *Additionally the group separator must be a line break (*\n*). Some RADIUS servers (e.g. MS NPS) will not
```
AdSchellevis, line 32: We add the separator when multiple `class` attributes are being received, if you can add these separators the behavior would be as you described, I just have no clue if that exists. 

https://github.com/opnsense/core/blob/41d8cf30c967d8d8336b37ae17b0592fc06c4bc6/src/opnsense/mvc/app/library/OPNsense/Auth/Radius.php#L524

adn77: This code would actually be much smoother:

* multiple groups would work
* no need to prfix with *CN=*

https://github.com/opnsense/core/issues/6230#issuecomment-1374377826

AdSchellevis: The "cn=" idea kind of started with the expectation that NPS (windows) passed these properties automatically, which later in time didn't appear to be the case. I have no clue if there are products that just pass `memberOf` to the class attributes, which would make more sense. 

As semicolon (;) isn't used in CN's as far as I know, I'm also fine with a new ticket in core to support semicolons as well for RADIUS, looking at the current code, we can support both without breaking existing things. 

If you open a ticket, I'll fix the code if you promise to test your use case and finish the documentation as well on this change.

adn77: I looked at some FreeRadius mailing list entries... 😄 Apparently a lot of magic can be done there populating the return attributes.

As I worked with other firewalls the important piece of information for me was the attribute name (in OPNsense it's "CLASS", Sophos XGS uses "FILTER-ID" as attribute) and the format of the group name ("CN=" in our case).

In case of NPS a workaround exists for defining different groups (just add multiple rules and return different CLASS attributes based on group matching in each restriction).

Therefore I would not recommend changing any code just now. I modified the explanation of the synchronize groups feature and removed the "single" mention.
Anybody working seriously with Radius should now be able to understand the inner workings from the Note: section.

AdSchellevis: it's magical indeed. Thanks for working on this, I'll merge this. As part of https://github.com/opnsense/core/issues/7904 we are going to change the user creation order, so you can create users without having groups assigned yet, but that's still under the hood.

PR #636, opened 2024-11-13
Update hardware.rst

Neozlag:
There is a mistake in state transition table calculation. The documention says : 
"1 table entry require 1kB, so 1000 entry require 10MB." But 1000*1kB = 1000kB = 1MB.


AdSchellevis:
@Neozlag thanks, that's an old one it seems :)

PR #641, opened 2024-11-21
net/ndproxy: Add ndproxy manual

Monviech:
For https://github.com/opnsense/plugins/pull/4348

### DIFF
--- a/source/manual/ndproxy.rst
+++ b/source/manual/ndproxy.rst
@@ -1,0 +1,226 @@
+==================================================
+ndproxy (Neighbour Discovery Proxy)
+==================================================
+
+.. contents::
+   :local:
+   :depth: 2
+
+This manual provides a quick overview of ndproxy and how to configure it for general use.
+
+
+Introduction to ndproxy
+==================================================
+
+Ndproxy is a kernel module that acts as a proxy for IPv6 Neighbor Discovery (ND) messages between a Provider Edge (PE) router
+and Customer Premises Equipment (CPE).
+
+When ndproxy runs on the same device as the CPE (e.g., OPNsense), it allows the device
+to act as both the home network’s router and the proxy for handling ND messages. This setup is particularly useful in cases
+where an ISP only provides limited IPv6 delegation (e.g., a single /64 prefix). By using ndproxy, such limitations can be
+bypassed to allow the LAN to use the ISP provided prefix.
+
+For more technical details: `ndproxy(4) <https://man.freebsd.org/cgi/man.cgi?query=ndproxy>`_
+
+
+Installation
+--------------------------------------------------
+
+Install ``os-ndproxy`` from :menuselection:`System --> Firmware --> Plugins`.
+
+
+Important configuration details
+--------------------------------------------------
+
+- **Promiscuous Mode**:
+    The listening interface (WAN) must be set to promiscuous mode. 
+    Otherwise the router can not join multicast groups to respond to solicitations for hosts in the LAN.
+  
+- **Link-Local Address**:
+    The listening interface (WAN) requires at least a link-local IPv6 address.
+    If you want to add a GUA (Global Unicast Address) to WAN, it must be with /128 Prefix.
+  
+.. Attention::
+
+   If WAN and LAN both have a GUA configured with the same /64 Prefix, this setup will not work. The network stack will not be able
+   to decide the sending interface, and routing will fail.
+   
+- **Switch configuration**:
+
+   If there is a switch between the PE and CPE router, ensure there is no MLD snooping or Multicast (IGMP) snooping configured
+   on the VLAN.
+
+
+Proxy /64 Prefix from WAN to LAN
+==================================================
+
+There must be a Provider Edge (PE) router that delegates an IPv6 Prefix to your OPNsense (CPE). You need to acquire the link-local IPv6 address of this PE router.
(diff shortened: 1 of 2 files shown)


source/manual/ndproxy.rst
```diff
@@ -0,0 +1,229 @@
...
+--------------------------------------------------
+
+Install ``os-ndproxy`` from :menuselection:`System --> Firmware --> Plugins`.
+
+
+Important configuration details
+--------------------------------------------------
+
+- **Promiscuous Mode**:
+    The listening interface (WAN) must be set to promiscuous mode.
```
fichtner, line 36: If the WAN is a VLAN you need to assign a parent and set promisc mode there. May be worth a note?

source/manual/ndproxy.rst
```diff
@@ -0,0 +1,229 @@
...
+
+Important configuration details
+--------------------------------------------------
+
+- **Promiscuous Mode**:
+    The listening interface (WAN) must be set to promiscuous mode. 
+    Otherwise the router can not join multicast groups to respond to solicitations for hosts in the LAN.
+  
+- **Link-Local Address**:
+    The listening interface (WAN) requires at least a link-local IPv6 address.
```
fichtner, line 40: This requirement is true but in 99% of cases that is a given (unless somebody wedges WAN on a bridge where link-local option was not set -- but the text here will not help those either)

source/manual/ndproxy.rst
```diff
@@ -0,0 +1,229 @@
...
+    If you want to add a GUA (Global Unicast Address) to WAN, it must be with /128 Prefix.
```
fichtner, line 41: Only if it is in the same /64 being proxied.

source/manual/ndproxy.rst
```diff
@@ -0,0 +1,229 @@
...
+  
+.. Attention::
+
+   If WAN and LAN both have a GUA configured with the same /64 Prefix, this setup will not work. The network stack will not be able
+   to decide the sending interface, and routing will fail.
+   
+- **Switch configuration**:
+
+   If there is a switch between the PE and CPE router, ensure there is no MLD snooping or Multicast (IGMP) snooping configured
```
fichtner, line 50: 👍

source/manual/ndproxy.rst
```diff
@@ -0,0 +1,229 @@
...
+        |                 | LAN                          WAN |                 |
+        |    Router PE    |----------------------------------|    Router CPE   |
+        |                 | fe80::1/64            fe80::2/64 |    (ndproxy)    |
+        +-----------------+ 2001:db8::1/64   2001:db8::2/128 +-----------------+
+                | WAN                                             LAN |
+                |                                                     |
+        Prefix Delegation: /56                                        |
+                |                                                     |
+          fe80::1/64                                            fe80::1/64
+          2001:db8::/56                                         2001:db8::3/64
```
fichtner, line 78: Not sure if 2001:db8::3/64 is strictly necessary, especially with a shifting prefix this is impractical. But in the dynamic case a track interface setting works anyway?

source/manual/ndproxy.rst
```diff
@@ -0,0 +1,229 @@
...
+**Promiscous Mode**                             ``X`` 
+                                                (important to respond to all NDP Multicasts)
+**Prefix Delegation Size**                      ``64``
+**Request Prefix Only**                         ``X``
+                                                (optional)
+==============================================  ====================================================================
+
+.. Note::
+
+   `Request Prefix Only` is optional, but needed if your WAN interface would autoconfigure a /64 GUA.
```
fichtner, line 153: Don't understand this. SLAAC is a separate range and /64. IA-NA is /128 -- IA-PD is /64 but not directly assigned to WAN so LAN could already use it?

fichtner (commented):
Maybe we should discuss the use cases before continuing the review. Looks like I'm missing something so better to clear that up first :)

Monviech:
> Maybe we should discuss the use cases before continuing the review. Looks like I'm missing something so better to clear that up first :)

Sounds like a good plan, I appreciate the help. :)

source/manual/ndproxy.rst
```diff
@@ -0,0 +1,229 @@
...
+**IPv6 Configuration Type**                     ``Static IPv6``
+**IPv6 address**                                ``2001:db8::1/64``
+==============================================  ====================================================================
+
+Go to :menuselection:`Services --> ISC DHCPv6 --> LAN`
+
+==============================================  ====================================================================
+**Enable**                                      ``X``
+**Range**                                       from: ``2001:db8::2`` to: ``2001:db8::2``
+**Prefix Delegation Range**                     from: ``2001:db8::`` to: ``2001:db8::``
```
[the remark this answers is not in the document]
Monviech: The GUA is not strictly needed for DHCPv6 to work here. The range can also be left empty.

But getting a /128 on the WAN can be important, e.g. for IPv6 VPN tunnels.

source/manual/ndproxy.rst
Monviech: Im not sure here either. It works with this enabled, and also disabled.

But if Router Advertisements would be set to Assisted on the Upstream Router, there could be a DHCPv6 /128 and a SLAAC /64 configured at the same time. I saw that happen, maybe it was a weird one time thing on my side but it wrecked my routing until I found that out.

fichtner (approved):
great, thanks for the redesign ❤️

Monviech:
Come on lets gooooooo

PR #643, opened 2024-12-04
source: change menu structure "other types" -> "devices" for clarity

fichtner:
For https://github.com/opnsense/core/issues/7446

TODO:

- [x] review current PPP pages (removed unreachable interface settings, but need to add a device page for it)
- [x] remove "use IPv4 connectivity" mentions
- [x] change PPPoE creation workflows (there is none that is explicit!)


Monviech:
Maybe you could deprecate the source/manual/other-interfaces.rst document by emptying it and saying

"This page moved to "LINK""

fichtner:
Could do, but to be frank this is simpler. I'm not overly attached to having the right document name. The content matters (same with static PHP pages in the GUI we have not renamed because it doesn't bring any noticeable benefit).

Monviech (approved):
Looks good to me

Issue #639, opened 2024-11-20
vpn/ipsec: Add additional information to swanctl roadwarrior docs


Eisaichen:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/docs/blob/master/CONTRIBUTING.md
- [x] I have searched the existing issues and I am convinced that mine is new.

**Is your feature request related to a problem? Please describe.**

On Windows, when split-tunnel is enabled, each time connected to VPN, the route has to be manually added by using the command provided by the document.
https://github.com/opnsense/docs/blob/a71715be5ce77606b6e85a04f4add0d45f330806/source/manual/how-tos/ipsec-swanctl-rw-ikev2-eap-mschapv2.rst?plain=1#L656
https://docs.opnsense.org/manual/how-tos/ipsec-swanctl-rw-ikev2-eap-mschapv2.html#client-configuration

**Describe the solution you like**

Windows can automate the process with the PowerShell command `Add-VpnConnectionRoute`, with no elevated privilege required. In this example with the document, the command will be:
``` powershell
Add-VpnConnectionRoute -ConnectionName 'vpn1.example.com' -DestinationPrefix '192.168.1.0/24' -PassThru

# IPv6
Add-VpnConnectionRoute -ConnectionName 'vpn1.example.com' -DestinationPrefix 'fe0d:abcd:1234:cafe::/64' -PassThru
```

Once connected, the configured route will add to the routing table automatically, and also delete if disconnected.
``` powershell
# Get corresponding route with VPN connection
(Get-VpnConnection -ConnectionName 'vpn1.example.com').routes

# Remove associate route
Remove-VpnConnectionRoute -ConnectionName 'vpn1.example.com' -DestinationPrefix '192.168.1.0/24' -PassThru
```

Monviech:
Thanks I didn't know this command existed. Want to offer a PR to add it to the document?

Eisaichen:
Hi, @Monviech 
I want to thank you and your team for this very detailed tutorial. I wish I could submit a PR however I lack confidence in my English for this kind of formal document.
In the meanwhile, despite following the document step by step, I still can not get my clients to work due to some minor issues. After a day of troubleshooting, I managed to solve those problems. I want to share what I found here as well.


### Option for Windows native client
(Tested on Win11 22631.4460)
[...]

PR #648, opened 2024-12-18
ipsec roadwarrior: add always send certificate to support native ios client

sebvonhelsinki:
## What
I've added a single line, setting the Send Certificate behavior to "Always".

## Why
Part of https://github.com/opnsense/docs/issues/639

We also stumbled about the Apple iOS related problem after following the current documentation, so I decided to improve it.

## Impact prognosis
Since the how-to sets up the authentication of the OPNSense side with a certificate, the certificate needs to be sent to the client during authentication. This change fixes the behavior for clients that don't request the cert without changing behavior for clients that already did request the cert.


Monviech:
Thank you :)

PR #650, opened 2025-01-02
Wireguard roadwarrior : Add tip to use peer generator and clarify DNS when using Unbound

remz1337:
Wireguard roadwarrior : Add tip to use peer generator and clarify DNS when using Unbound. The current tuto doesn't mention the peer generator feature. I think the right way would be to rewrite section 2 entirely to leverage the peer generator feature and explain to use the QR code before saving it, but at least with the tip people will know they can use it (and configuration is pretty much the same).

Also, client config needs the DNS to be the tunnel address for it to work.

=====
For context, I've been trying to setup VPN for the past few days. I gave up on IPSec and tried Wireguard next. Followed the how-to step by step a few times and never could get it to work, until I found that I simply needed to configure the DNS on the client side to be the address of the tunnel (10.10.10.1 in the tuto example).

### DIFF
--- a/source/manual/how-tos/wireguard-client.rst
+++ b/source/manual/how-tos/wireguard-client.rst
@@ -50,7 +50,7 @@ Step 2 - Configure the client peer
 
 .. Tip::
 
-    Peers can be generated using the new peer generator feature under :menuselection:`VPN --> WireGuard --> Peer generator`. If using the peer generator and Unbound DNS, fill the DNS server with the tunnel address (eg* :code:`10.10.10.1` *).
+    Peers can be generated using the new peer generator feature under :menuselection:`VPN --> WireGuard --> Peer generator`. If using the peer generator and require Unbound DNS to serve names, fill the DNS server with the tunnel address (eg* :code:`10.10.10.1` *).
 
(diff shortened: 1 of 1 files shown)


source/manual/how-tos/wireguard-client.rst
```diff
@@ -48,6 +48,10 @@ Step 1 - Configure the Wireguard Instance
 Step 2 - Configure the client peer
 ---------------------------------------------
 
+.. Tip::
+
+    Peers can be generated using the new peer generator feature under :menuselection:`VPN --> WireGuard --> Peer generator`. If using the peer generator and Unbound DNS, fill the DNS server with the tunnel address (eg* :code:`10.10.10.1` *).
```
AdSchellevis, line 53:
```suggestion
    Peers can be generated using the new peer generator feature under :menuselection:`VPN --> WireGuard --> Peer generator`. If using the peer generator and require Unbound DNS to serve names, fill the DNS server with the tunnel address (eg* :code:`10.10.10.1` *).
```

Monviech:
Thank you for your contribution :)

PR #654, opened 2025-01-07
Update settingsmenu.rst: table formating

sjjh:
Some tables use a syntax where no trailing "|" is needed. Removed it for the rows in these tables.


AdSchellevis:
@sjjh thanks!

Issue #668, opened 2025-02-07
Interfaces: Devices: Point to Point section needs a mention in the devices manual page


fichtner:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/docs/blob/master/CONTRIBUTING.md
- [x] I have searched the existing issues and I am convinced that mine is new.

**Is your feature request related to a problem? Please describe.**

The removal of 9ed75e3e5c05 showed that we don't actually have a page describing the PPP settings. The access of PPP settings through the interfaces settings has been removed to make the code more maintainable.

**Describe the solution you like**

- [x] Add a Point to Point section with the most useful settings removed in 9ed75e3e5c05 plus some more as these could be important to users.
- [x] Also add a tutorial for PPPoE setup from scratch via the GUI (not the wizard) which is something we are definitely missing as well.

**Describe alternatives you considered**

N/A

**Additional context**

#643

wrobelda:
I just run into this issue, somewhat. The docs mention PPP/Wireless log access in several places, but this is nowhere to be found in the UI anymore. See e.g.:
https://docs.opnsense.org/manual/how-tos/cellular.html#step-4-troubleshooting ("Interfaces ‣ Point-to-Point ‣ Log File")
https://docs.opnsense.org/manual/logging_interfaces.html

fichtner:
I think I already deleted one, but not the other. Logging of connectivity moved to where it belongs—the general log along with  lall other connectivity daemons and reconfiguration logging.

PR #674, opened 2025-02-19
devices: Add Point-to-point to general overview and add how-to for ISP setup with PPPoE

Monviech:
Fixes: https://github.com/opnsense/docs/issues/668

### DIFF
--- a/source/manual/other-interfaces.rst
+++ b/source/manual/other-interfaces.rst
@@ -372,4 +372,6 @@ The following `Link Types` are available for these interfaces:
       .. Note:: Check the PPTP and PPP tab for all options.
 
 
+.. Note:: The ``ppp`` log files can be found in :menuselection:`System --> Log Files --> General`.
+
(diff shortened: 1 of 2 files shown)


source/manual/other-interfaces.rst
```diff
@@ -265,3 +265,111 @@ source in the encapsulating IPv4/IPv6 header.
...
+      **Username**                              Enter your username for CHAP authentication.
+      **Password**                              Enter your password for CHAP authentication.
+      ========================================= ====================================================================================
+
+      .. Note:: Check the PPP tab for all advanced options.
+
+
+   .. tab:: L2TP (Layer 2 Tunneling Protocol)
+
+      An extension of PPTP that does not provide encryption on its own but is often paired with IPsec
```
fichtner, line 369:
```suggestion
      An extension of PPTP that does not provide encryption on its own but was often paired with IPsec
```

;)

PR #682, opened 2025-03-28
Update wireguard-selective-routing.rst

Bishop-trevorstuart:
Added additional note for Step 2 to clarify for users of VPN providers that give the private key and listen ports to use.

### DIFF
--- a/source/manual/how-tos/wireguard-selective-routing.rst
+++ b/source/manual/how-tos/wireguard-selective-routing.rst
@@ -50,9 +50,9 @@ Step 2 - Configure the WireGuard Instance
     ===================== ===============================================================================================
      **Enabled**           *Checked*
      **Name**              *Call it whatever you want (eg* :code:`VPNProviderName` *)*
-     **Public Key**        *This will initially be blank; it will be populated once the configuration is saved*
-     **Private Key**       *This will initially be blank; it will be populated once the configuration is saved*
-     **Listen Port**       *51820 or a higher numbered unique port*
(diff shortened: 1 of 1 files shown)


Monviech:
Getting your private key from the VPN provider sounds wild as it must be secret.

Bishop-trevorstuart:
If it's bad info then no issue at all. When I was following these instructions my WG peer would not connect until I opened the wg0 instance and put the private key Torguard included in the config file I generated from them, and the port they provided in same config. It's very possible(LIKELY) I just don't understand something of course. But that was my experience and I thought maybe other VPN providers were simular that you log into your account and generate the config from their site. 
For reference here is what the config generator looks like, you do have to provide your VPN accounts username and to generate the config I leave those private and public keys empty. 
image

And here is the conf file that is generated (All the details including keys are sanitized). 
image

Based on the instructions provided here I was beating my head against a wall, then thought maybe I'm supposed to use that private key they provided me... It connected and the selective routing is working as desired now.

Monviech:
The issue is that you should give them your public key, and the provider should give you their public key.

The public key is the one that is shared. The private key is private, hence the name. It is generated and kept secret by each side.

If the provider generates both private and public key pairs for both wireguard peers, its kinda insecure by design.

Anybody who has your private key can decrypt the data that is sent to you. I know that its just your VPN provider and you trust them unconditionally since all data will go between your box and the vpn provider anyway, but the key could be accidentally shared and then a third party can capture the traffic sent to you and decrypt it.

Monviech:
@Bishop-trevorstuart Thank you for the effort, even though it was not merged. Much appreciated :)

PR #691, opened 2025-04-05
wireguard-s2s-howto: Keepalive interval requirement

philtlp:
Added a note emphasizing the importance of configuring the **Keepalive interval** for at least one peer when setting up a wireguard Site-to-Site VPN with two OPNsense firewalls. Without this configuration, the connection will not be initiated, and the VPN will fail to establish.

### DIFF
--- a/source/manual/how-tos/wireguard-s2s.rst
+++ b/source/manual/how-tos/wireguard-s2s.rst
@@ -74,7 +74,7 @@ Go to tab **Peers** and press **+** to create a new peer.
 
 Enable the *advanced mode* toggle.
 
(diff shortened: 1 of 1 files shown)


Monviech:
I'm interested about the following:
- Was either OPNsense behind a NAT itself?
- Were there missing inbound firewall rules? (The keepalives on both sides might have created states for the return traffic)
- Did you try initiating a ping from both sides at the same time to a target on the other side, and see if it had the same effect as setting the keepalives?

I have created quite a lot of wireguard tunnels between firewalls and having the keepalive as requirement if no NAT traversal or states must be first established is curious.

PR #693, opened 2025-04-09
pppoe: Correct isp setup guide

Monviech:
I read some feedback on the forum and reddit with people struggling with this guide, so I double checked and corrected the mistake in it.

### DIFF
--- a/source/manual/how-tos/pppoe_isp_setup.rst
+++ b/source/manual/how-tos/pppoe_isp_setup.rst
@@ -54,7 +54,11 @@ Option                              Value
 
 - Press **Save** and **Apply**
 
-.. Tip:: If you do not plan to use this interface at all, you can also unassign it. PPPoE does not require this interface to be assigned.
+.. Attention::
+
+    The parent interface should stay assigned in :menuselection:`Interfaces --> Assignments` to correctly spoof the mac address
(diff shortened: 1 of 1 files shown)


source/manual/how-tos/pppoe_isp_setup.rst
```diff
@@ -52,15 +54,15 @@ Option                              Value
 
 - Press **Save** and **Apply**
 
-.. Tip:: If you do not plan to use this interface at all, you can also unassign it. Only the VLAN interface must be assigned for PPPoE to work.
+.. Tip:: If you do not plan to use this interface at all, you can also unassign it. PPPoE does not require this interface to be assigned.
```
fichtner, line 57: If ISPs and modems are involved it may be worth mentioning that the parent has to be assigned to correctly spoof the mac address on the vlan, because only spoofing the vlan will discard the traffic in the parent interface if not set to promisc mode (which also requires assignment of the parent)

source/manual/how-tos/pppoe_isp_setup.rst
```diff
@@ -52,15 +54,15 @@ Option                              Value
...
 
 
-2. VLAN Interface
+2. VLAN Device
```
fichtner, line 60: good idea :)

fichtner (approved):

fichtner:
thanks 👍

Issue #705, opened 2025-04-30
shaper: Create documentation and explanation for Control plane shaping


SeimusS:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [X] I have read the contributing guide lines at https://github.com/opnsense/docs/blob/master/CONTRIBUTING.md
- [X] I have searched the existing issues and I am convinced that mine is new.

**Is your feature request related to a problem? Please describe.**

A lot of users using Shapers, do not separate control plane of L3 protocols. This is causing network disruption.
Example of this issue is seen even for users not using dynamic routing protocols but running IPv6. Where during a congestion event back-pressure of the queues causes ICPMv6 dropping which results in IPv6 disruption.

Examples of reported problems https://github.com/opnsense/core/issues/7342 https://github.com/opnsense/core/issues/6714

This issues are more visible when users use schedulers that use AQM or measure sojourn time for a packet in a queue such as FQ_CoDel. This is due to the fact, even if packet is queued but its time in a queue is to much it will start TAIL dropping.

**Describe the solution you like**

- Introduce proper Concepts of the planes in networking
- Explanation of the issue and necessity so separate control plane from rest of the traffic
- Create an example of such Shaper
- Update doc "Fighting Bufferbloat with FQ_CoDel" with a warning and pointing to the docs for use separate class/Pipe + Queues for control plane

**Describe alternatives you considered**

Topic on forum is already created, but a proper refined official doc would be more beneficial
https://forum.opnsense.org/index.php?topic=46990.0

**Additional context**

I will contribute to this. Feature request is open just to indicate there is a need and intent to create such specific documentation.

PR #706, opened 2025-04-30
shaper: Control plane shaping

SeimusS:
Fixes 
https://github.com/opnsense/docs/issues/705 
https://github.com/opnsense/core/issues/6714 
https://github.com/opnsense/core/issues/7342

### DIFF
--- a/source/manual/how-tos/shaper_control_plane.rst
+++ b/source/manual/how-tos/shaper_control_plane.rst
@@ -212,7 +212,6 @@ Step 4 - Finalizing the configuration
 
 Now press **Apply** to activate the traffic shaping rules.
 
-.. |apply| image:: images/applybtn.png
 
 -----------------------
 


SeimusS:
Initial contribution from forum user OPNenthu.
I am not sure what is his username on github.

source/manual/how-tos/shaper_control_plane.rst
```diff
@@ -0,0 +1,233 @@
...
+On the **Queues** tab click the **+** button in the lower right corner.
+An empty **Edit queue** screen pops up.
+
+Follow the same process as for the Download queue, 
+selecting the **"Control-plane-Pipe-Upload**,
+and entering "Control-plane-IPv6-Queue-Upload" for the **description**
+
+Step 3a - Create Download Rule
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+On the **Rules** tab click the **+** button in the lower right corner.
```
swhite2, line 174: fwiw, in the long run, `ipfw` rules might not be the preferred choice anymore to do this. In theory a separate rule in `pf` for i.e. ipv6-icmp should achieve the same (verifiying this would be even better).

dstapa:
> Initial contribution from forum user OPNenthu. I am not sure what is his username on github.

@SeimusS , thanks for the mention (OPNenthu here)...

I think the Control plane docs are clear and easy to follow.

My only suggestion would be to consider adding a warning in shaper_bufferbloat.rst about the need to do this if using FQ_CoDel with IPv6.  The fact that CoDel/AQM has been observed to break IPv6 connectivity for some users should be explicitly stated in that document, IMO.

source/manual/how-tos/shaper_control_plane.rst
SeimusS: I am not going to deny that I didn't consider it. But isn't that feature considered yet Experimental?

SeimusS:
@dstapa hey there,
 yes I agree and its already there mentioned within this PR ;)

`.. Attention::
    If you are running IPv6 or any dynamic routing protocol, consider creating a Control plane specific class how-tos/shaper_control_plane.


`

source/manual/how-tos/shaper_control_plane.rst
swhite2: It is, although adoption of pf-based shaping would be good to eventually get it out of that state, so maybe an honorable mention or "tip"?

SeimusS: Definitely at minimum a honorable mention would be on place which is not problem to add.

But I think it would be even better to to have it as optional approach with example

(Optional) Step 3c - Create Download Rule
(Optional) Step 3d - Create Upload Rule

* However here comes the question how to do it properly considering the default inbuilt OPNsense IPv6 rules?
* Do we need in such case only one rule in one direction specified on WAN, example IN rule?
* And than specify Queues in Traffic shaping/rule direction & Traffic shaping/reverse direction within the same rule? Example:
         Traffic Shaping (rule direction): Control-plane-IPv6-Queue-Download
         Traffic Shaping (reverse direction): Control-plane-IPv6-Queue-Upload
* Or separate in out rule is needed with specific shaper direction?
* Will be those rules even hit, as the default IPv6 ICMPv6 rules are quick and floating?

source/manual/how-tos/shaper_control_plane.rst
```diff
@@ -0,0 +1,233 @@
...
+   
+
+
+
+Necessity of Control Plane Shaping
+----------------------------------
+
+Control plane is a necessary component for the network operation.
+A congested state of the network, if control plane is not taken care of, will result in network disruptions. 
+In order to guarantee operation and non-disruption in such events, we can take advantage of QoS in our case Shaping.
```
swhite2, line 44:
```suggestion
In order to guarantee operation and avoid disruption in such events, we can take advantage of Traffic Shaping
```

source/manual/how-tos/shaper_control_plane.rst
```diff
@@ -0,0 +1,233 @@
...
+
+**Its necessary to understand Control plane as such, needs to be always taken care of, always in its own dedicated way.**
```
swhite2, line 46: (following up on previous sentence)

```suggestion
**With Shaping, we can reserve dedicated bandwidth for the control plane**
```

source/manual/how-tos/shaper_control_plane.rst
```diff
@@ -0,0 +1,233 @@
...
+
+.. Attention::
+    Control plane should not share a class (Pipe nor Queue) with any other traffic type.
```
swhite2, line 49:
```suggestion
    The control plane should not share a class (Pipe or Queue) with any other traffic type.
```

source/manual/how-tos/shaper_control_plane.rst
```diff
@@ -0,0 +1,233 @@
...
+
+
+Configuring Control plane for OPNsense
+--------------------------------------
+
+Configuration will take in account existence of already configured shaper as per how-tos/shaper_bufferbloat
```
swhite2, line 55:
```suggestion
The configuration will take into account the existence of an already configured shaper as per :doc:`how-tos/shaper_bufferbloat`
```

(please test the above link when compiled)

source/manual/how-tos/shaper_bufferbloat.rst
```diff
@@ -37,6 +37,10 @@ FQ_CoDel begins to mark or drop some of those packets to cause the sender to slo
 
 For more details, see RFC 8290 https://datatracker.ietf.org/doc/html/rfc8290. 
 
+.. Attention::
+    If you are running IPv6 or any dynamic routing protocol, consider creating a Control plane specific class how-tos/shaper_control_plane.
```
swhite2, line 41:
```suggestion
    If you are running IPv6 or any dynamic routing protocol, consider creating a Control plane specific class :doc:`how-tos/shaper_control_plane`.
```

source/manual/how-tos/shaper_control_plane.rst
```diff
@@ -0,0 +1,233 @@
...
++----------------+----------+-----------------+
+|    1% of BW    | Download |      Upload     |
++================+==========+=================+
+|      Mbit/s    |    5     |       0.5       |
++----------------+----------+-----------------+
+
+The 1% is the default lowest value. Its subjected to be increased if needed.
+
+.. Attention::
+    BW given to a different Pipe needs to be subtracted from already existing one.
```
swhite2, line 77:
```suggestion
    BW given to a different Pipe needs to be subtracted from already existing ones.
```

source/manual/how-tos/shaper_control_plane.rst
```diff
@@ -0,0 +1,233 @@
...
+ **Enable CoDel**         (empty)                           *Leave empty for a queue*
+ **(FQ-)CoDel target**    (empty)                           *Leave empty for a queue*
+ **(FQ-)CoDel interval**  (empty)                           *Leave empty for a queue*
+ **(FQ-)CoDel ECN**       (empty)                           *Leave empty for a queue*
+ **description**          Control-plane-IPv6-Queue-Download *Free field, enter something descriptive*
+========================= ================================= ==============================================================================
+
+.. Note::
+
+        Using WFQ or QFQ and weights, we can tell how much BW will get each control plane for each protocol gain.
```
swhite2, line 161:
```suggestion
        Using WFQ or QFQ and weights, we can tell how much BW each protocol control plane will get.
```

swhite2 (changes requested):

source/manual/how-tos/shaper_control_plane.rst
```diff
@@ -0,0 +1,233 @@
...
+of a network device.
+
+In essence the concept creates logical layers into which are categorized packets 
+per their use and type.
+
+There are 3 Planes:
+
+1. **The control plane**
+  
+   Affect the operationability of the network.
```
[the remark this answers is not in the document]
SeimusS: Thank you! I this word specifically broke my mind

source/manual/how-tos/shaper_control_plane.rst
```diff
@@ -0,0 +1,233 @@
...
+|      Mbit/s    |   495    |        30       |
++----------------+----------+-----------------+
+
+
+Bandwidth allocation requirements for Control plane:
+
++----------------+----------+-----------------+
+|    1% of BW    | Download |      Upload     |
++================+==========+=================+
+|      Mbit/s    |    5     |       0.5       |
```
dstapa, line 71: Cannot enter 0.5 Mbit/s into the pipe BW field.  Maybe change the example values, or the units to kbit/s?

source/manual/how-tos/shaper_control_plane.rst
swhite2: > However here comes the question how to do it properly considering the default inbuilt OPNsense IPv6 rules?

Good point, someting we'll have to think about, but the use case hadn't popped up yet.

> Do we need in such case only one rule in one direction specified on WAN, example IN rule?

Since the pf rules are stateful by default, the "opposite" direction option (in the rule edit) is used to replicate the behavior of a second rule in `ipfw`. This can be a queue or a pipe, if a queue is selected, the associated pipe is used as well.

SeimusS: > Good point, someting we'll have to think about, but the use case hadn't popped up yet.

Hehe, looks like now we have a potential use case, and actually one that could be one most the most common.

One way how to do it, would be to remove those default rules and manually set them. But..
A. too intrusive
B. not a good approach
C. will cause a lot of problems for users


> Since the pf rules are stateful by default, the "opposite" direction option (in the rule edit) is used to replicate the behavior of a second rule in `ipfw`. This can be a queue or a pipe, if a queue is selected, the associated pipe is used as well.

Aaaah interesting. 

If the Traffic Shaping field is populated in the pf rule, the rule is only used to classify traffic into a Pipe or Queue?
Or does it as well perform deny,refuse,permit actions as a rule without Traffic Shaping field?

source/manual/how-tos/shaper_control_plane.rst
SeimusS: done, thanks for the input!

source/manual/how-tos/shaper_control_plane.rst
swhite2: > If the Traffic Shaping field is populated in the pf rule, the rule is only used to classify traffic into a Pipe or Queue?
>  Or does it as well perform deny,refuse,permit actions as a rule without Traffic Shaping field?

It does everything as it normally would according to the pf rule. I don't think the packet is in any way touched by shaping if it is blocked. Therefore, you'd see pipes/queues configured on pass rules.

SeimusS: > > However here comes the question how to do it properly considering the default inbuilt OPNsense IPv6 rules?
> 
> Good point, someting we'll have to think about, but the use case hadn't popped up yet.

Thinking about this,

Ideas:
1. The only possible way from user perspective, would be to recreate those rules as floating quick and set them above the default rules.

2. Currently the default rules are non editable, what if they were? In that case a user could just set in the proper Queue or Pipe.

3. Maybe a bit wishful, but what if OPNsense came out of the box with pre-configured Control Plane Shaper/class into which the default rules would be already configured?

SeimusS:
I know there is still an open topic about using the pf's Traffic Shaping section to create rules for the pipe/queue. But for the IPv6 example it introduces a bit of a challenge as the default IPv6 rule-set would interfere with them.

Another point tho, the rules set in this doc example, match any IPv6 ICMP type any IP. Thus it will match as well ICMP from hosts not only from the GW/Router. Usually we want to match traffic that is control plane for the network itself, respectively Source or Destination depending on the Direction should be set to the GWs/Routers IP.

Other than that I think this doc is ready, and I think it could be pushed. We can always add the "optional" way for rules later if we somehow figure out how to do it properly.

Thus I am flagging this PR ready for review.

source/manual/how-tos/shaper_control_plane.rst
```diff
@@ -0,0 +1,233 @@
...
+- **description** (set to "Control-plane-IPv6-Rule-Upload")
+
+.. Attention::
+    The sequences needs to be at least 1 less than the default any rules or rules that would catch this traffic type 
+    and you may need to adjust the other rule sequence values accordingly.
+
+Step 4 - Finalizing the configuration
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Now press |apply| to activate the traffic shaping rules.
```
Monviech, line 213:
```suggestion
Now press **Apply** to activate the traffic shaping rules.
```

swhite2 (approved):

swhite2:
@SeimusS Many thanks for your work :)

Monviech:
@SeimusS Thank you a lot :)

source/manual/how-tos/shaper_control_plane.rst
SeimusS: @swhite2 @Monviech 
The docs got compiled, and I checked if the link works but is not, basically its not even represented as a link.

Is this something you guys can quickly fix? :)

swhite2: @SeimusS Should be https://github.com/opnsense/docs/commit/284f78ffc48f134c3675cbe1bf82189d7f9f7ce4, thanks for checking :)

PR #708, opened 2025-05-01
boot: update boot troubleshooting guide

swhite2:
should we also do

```
set hint.uart.1.at="isa"
```

?


AdSchellevis (approved):

swhite2:
Ignoring `set hint.uart.1.at="isa"` for now.

PR #709, opened 2025-05-03
Type-hint ApiParser using TypedDict

fsackur:
Hi, I'm looking to build on this code and I would personally find type-hints helpful. Feel free to reject if you don't want them. In that case, I'll remove from my own branch before any future PR.

The problem to be solved: when consuming the library, a developer must manually track the expected keys to avoid KeyError. With type hints, the IDE can auto-complete the correct keys and hint the values.

As you probably know, TypedDict is a check-time construct and has no runtime effect.

My personal preference is to use pydantic, as the syntax is cleaner, but it would add a dependency and has a runtime effect. If you'd like me to resubmit using pydantic, let me know, and I'll update and test.

### DIFF
--- a/lib/__init__.py
+++ b/lib/__init__.py
@@ -1,8 +1,29 @@
 import os
 import re
-from lib.phply.phpast import Class,  MethodCall, Constant, UnaryOp
+from typing import List, Literal, NotRequired, TypedDict
+from lib.phply.phpast import Class, MethodCall, Constant, UnaryOp
 from lib.phply.phplex import lexer
 from lib.phply.phpparse import make_parser
+
+
+HttpMethod = Literal["GET", "POST", "GET,POST"]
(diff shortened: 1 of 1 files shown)


AdSchellevis:
@fsackur I haven't really used them earlier, but don't mind adding them either. So far it hasn't drawn my attention much as these constructs suggests a strictness that doesn't exist runtime, which is less practical when building smaller scripts and not using an IDE, in which case it's easy to forget about updating the metadata as no visible signs will emerge.

For some reason using `pydantic` does sound a bit tempting as that would solve my primary concern, not sure about the dependencies, but for the docs rendering the FreeBSD part is a bit less relevant (I'm building these on osx anyway)

fsackur:
Duplicate of #713 

Closes #713

PR #711, opened 2025-05-03
Note about menu cache file

### LINKED ISSUE #710: Menu system does not automatically pick up changes before cache TTL expires
**Important notices**

- [x] I have read the contributing guide lines at https://github.com/opnsense/docs/blob/master/CONTRIBUTING.md
- [x] I have searched the existing issues and I am convinced that mine is new.

**Describe the issue**

https://github.com/opnsense/docs/blob/5382acd/source/development/examples/helloworld.rst?plain=1#L858-L859

The development doc states "the menu system automatically picks up this new information". This is not entirely true. There is a caching mechanism, defined in https://github.com/opnsense/core/blob/d0c9ace/src/opnsense/mvc/app/models/OPNsense/Base/Menu/MenuSystem.php.

**Suggestions**

Change paragraph to identify where the cache file is defined and add a note to the user to delete it.

**Version affected**

25.1

N/A

**Additional context**

N/A

fsackur:
Fixes #710 

I also believe, from reading the code, that a similar discrepancy would occur in the ACL system, because that also implements a cache. I have not touched ACLs myself. I offer a second commit in this PR in case the maintainers feel it would be useful.

### DIFF
--- a/source/development/examples/helloworld.rst
+++ b/source/development/examples/helloworld.rst
@@ -892,6 +892,15 @@ This creates an ACL key named “page-user-helloworld” which authorizes
 access to both the ui and API urls of this application. You can now
 grant access to this module from the system user manager.
 
+The ACL system is subject to caching, so you may not see your changes
+in the user manager page yet. Delete the
+``/tmp/opnsense_acl_cache.json`` file, if it exists. Now, when you
+refresh the user manager page, you should see that the new ACL is
+available to be assigned.
+
+The cache filename is defined in ``ACL.php``. As of v25.1, it is
+``/tmp/opnsense_acl_cache.json``.
+
 
 ----------------------------
 Create an installable plugin
--- a/source/development/examples/helloworld.rst
+++ b/source/development/examples/helloworld.rst
@@ -859,9 +859,6 @@ in the UI yet. Delete the ``/tmp/opnsense_menu_cache.xml`` file, if it
 exists. Now, when you refresh your page, you should notice the menu
 system automatically picks up the new information.
 
-The cache filename is defined in ``MenuSystem.php``. As of v25.1, it is
-``/tmp/opnsense_menu_cache.xml``.
-
 |menu registration|
 
 ------------------------------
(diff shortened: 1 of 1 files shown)


source/development/examples/helloworld.rst
```diff
@@ -854,9 +854,13 @@ adding this content into the Menu.xml:
...
     </menu>
 
+The menu system is subject to caching, so you may not see your changes
+in the UI yet. Delete the ``/tmp/opnsense_menu_cache.xml`` file, if it
+exists. Now, when you refresh your page, you should notice the menu
+system automatically picks up the new information.
 
-When you refresh your page now, you should notice the menu system
-automatically picks up this new information.
+The cache filename is defined in ``MenuSystem.php``. As of v25.1, it is
```
AdSchellevis, line 862: probably better to remove version references, the mechanism already exist for roughly 8 years ;) (https://github.com/opnsense/core/commit/b5d52b8a759b12d137e087b45ea6db82deb14f87)

source/development/examples/helloworld.rst
```diff
@@ -888,6 +892,15 @@ This creates an ACL key named “page-user-helloworld” which authorizes
 access to both the ui and API urls of this application. You can now
 grant access to this module from the system user manager.
 
+The ACL system is subject to caching, so you may not see your changes
+in the user manager page yet. Delete the
+``/tmp/opnsense_acl_cache.json`` file, if it exists. Now, when you
+refresh the user manager page, you should see that the new ACL is
+available to be assigned.
+
+The cache filename is defined in ``ACL.php``. As of v25.1, it is
```
AdSchellevis, line 901: same here

AdSchellevis:
@fsackur except the version reference, I think this is a good addition, after reading the text again, in these cases it's likely better to explain internals (which is in your text) then tooling to flush (which was my first thought). Thanks!

fsackur:
Done. On balance, I thought better to not mention where the internals are defined, because it's a) easy to grep for, and b) makes the dev doc less concise. I can add it back if you prefer.

source/development/examples/helloworld.rst
fsackur: Removed

source/development/examples/helloworld.rst
fsackur: Removed

AdSchellevis:
@fsackur looks good, thanks. I agree about the reference, its easy to find when you would like to know about the internals.

PR #712, opened 2025-05-03
collect_api_endpoints.py: factor out parse and render functions

fsackur:
Hi, I'm trying to re-use @AdSchellevis ' code for parsing, and in this file, I'd like to re-use the code that walks the source tree.

I don't have a strong need for this. Feel free to close, it's easy to reimplement.

If you happen to merge  #709 before this, then let's also include https://github.com/fsackur/opnsense_docs/commit/88cc00b

### DIFF
--- a/collect_api_endpoints.py
+++ b/collect_api_endpoints.py
@@ -39,16 +39,11 @@ def source_url(repo, src_filename):
     else:
         return "https://github.com/opnsense/core/blob/master/%s" % "/".join(parts[parts.index('src'):])
 
-if __name__ == '__main__':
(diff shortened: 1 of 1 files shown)


AdSchellevis:
@fsackur I've moved the "api walk" function into its own module in https://github.com/opnsense/docs/commit/db14cb93d607bf02c5096aa39e6dc643a35165a2, but I'm not sure if it does make much sense to isolate render in this case as it not really capable of rendering something else than the main purpose of this function.

fsackur:
Sounds like we don't need this PR

PR #713, opened 2025-05-04
Type-hint ApiParser using pydantic

fsackur:
Alternative approach using pydantic instead of TypedDict. It's a larger delta, but dot-syntax is nicer than `['key']` IMO.

I have verified that there's no diff in .rst output from collect_api_endpoints compared to master.

For awareness, each instantiation performs validation, so runtime errors may occur that would otherwise slip into Jinja and perhaps not cause an issue. IMO, this is an advantage.

Duplicate of #709 

Closes #709

### DIFF
--- a/collect_api_endpoints.py
+++ b/collect_api_endpoints.py
@@ -59,22 +59,22 @@ if __name__ == '__main__':
         }
         for controller in all_modules[module_name]:
             payload = {
-                'type': controller['type'],
-                'module': controller['module'],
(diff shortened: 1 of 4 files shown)


AdSchellevis:
@fsackur thanks, I've added https://github.com/opnsense/docs/commit/ed4048d011d03f3c32225fddbd6b1cfeebf3c0ec on top to clarify the method which should make reuse easier as well.

PR #714, opened 2025-05-04
Housekeeping: ignore generated parsetab.py

fsackur:
I see this changing when we run collect_api_endpoints. Is it helpful to ignore it?

Feel free to close without comment - I don't have the full context and I can use a global ignore.


AdSchellevis:
@fsackur I forgot about this one, thanks!

PR #717, opened 2025-05-06
ndproxy - add cloud example and refer to it in ipv6 document

AdSchellevis:
Add an example where ndproxy can be of use

### DIFF
--- a/source/manual/ipv6.rst
+++ b/source/manual/ipv6.rst
@@ -146,7 +146,8 @@ received. This part of the configuration is specified on the lan interface as "t
 .. Note::
 
     When your provider or cloud hosting does not offer a PD, routing is not reliably possible. In some cases
-    network address translation may be an option. The is an experimental `rfc <https://datatracker.ietf.org/doc/html/rfc4389>`__
+    network address translation may be an option (example in  :doc:`ndproxy <ndproxy>`).
+    There is an experimental `rfc <https://datatracker.ietf.org/doc/html/rfc4389>`__
     to proxy neighbor discovery packets, but no implementation exists yet.
 
 .. Note::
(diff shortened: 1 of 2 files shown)


Monviech:
@AdSchellevis I've tested this but I had to change two things to make it work:

- Use the WAN MAC in "Downlink MAC Address"
- Set WAN interface to either promiscuous or allmulti mode

Afterwards my client with the ULA could ping the internet.

All in all this setup seems pretty promising as it eliminates most of the pain points of using the GUA on the LAN side. Thank you.

AdSchellevis:
@Monviech it might be the case that my wan was in promisc as I installed the package from the repo on our end. just to be sure, did you reboot the machine (clean state) to validate if the lan mac doesn't work? In my setup it didn't make much of a difference, but I may have to check the source to understand what they actually try to do there.

AdSchellevis:
I think it should be wan in our case indeed as we try to update the ND database from our wan end, ref https://github.com/AlexandreFenyo/ndproxy/blob/dd93a2498569dff70e5c2654f51cc196e64d8300/ndpacket.c#L363

According to the source, they need promisc in order for the proxy to work

https://github.com/AlexandreFenyo/ndproxy/blob/dd93a2498569dff70e5c2654f51cc196e64d8300/ndpacket.c#L52-L58

In theory we should only need a limited "proxy" feature, which registers the addresses we use on our end (actually the mapped ones from the local net), `ndp` offers a proxy flag for these cases, but not a full proxy option.

I'll update the doc and retest it on my end. Still not a huge fan of this stuff, the code doesn't look very well organized in my humble opinion, but having a option which fits sane architectural choices beats having none in suppose.

Monviech:
It would still be pretty cool if the code would only join the needed multicast groups selectively, which I think would remove the need for promisc or allmulti.

The kernel should be able to handle this natively.

AdSchellevis:
To some degree it would make sense if the kernel could handle this, but realistically, this isn't very different from how `arp` works on IPv4, it needs a proxy too in order to answer to off link requests (https://man.freebsd.org/cgi/man.cgi?query=choparp)

Searching google a bit, seems to suggest it's roughly the same on linux (https://github.com/setaou/ndp-proxy)

PR #722, opened 2025-05-08
captive portal: update documentation

### LINKED ISSUE #676: Captive Portal: update / revamp docs
**Important notices**

- [x] I have read the contributing guide lines at https://github.com/opnsense/docs/blob/master/CONTRIBUTING.md
- [x] I have searched the existing issues and I am convinced that mine is new.

**Is your feature request related to a problem? Please describe.**

Captive Portal has been migrated to `pf` (https://github.com/opnsense/core/issues/8326), there are very little functional changes here except for the automatic rules that were added:

- DNS
- Access to the captive portal itself
- Blocking any traffic that isn't authenticated

Like before, no pass rules are generated automatically for normal network/outside access, therefore an explicit pass rule is still necessary to allow traffic to the internet/other networks. These rules do not need to account for the captive portal zones, these are filtered already with a higher priority.

Support for RFC 8908 (standardized captive portal status) has also been added https://github.com/opnsense/core/issues/8261, provided we can provision the URI for clients over DHCP (and RA/DHCPv6 at a later point), initial support for this is added in dnsmasq (https://github.com/opnsense/core/commit/eb2af7f2fd7bd2c79d927adb438a05a48802cf5e), this requires the administrator to manually set the URL in the DHCP option. Currently, this is `https://<opnsense-hostname>:<8000 + captive portal zone id>/api/captiveportal/access/api` (https://github.com/opnsense/core/blob/master/src/opnsense/mvc/app/controllers/OPNsense/CaptivePortal/Api/AccessController.php#L300).

RFC 8908 is a probe that does not involve user action, but enforces HTTPS and must reject any invalid certificate. Therefore, if an admin wants to make use of this option a self-signed certificate cannot be used on the captive portal unless the client trusts this certificate, or make use of e.g. a wildcard certificate. The RFC further states that a client needs to be able to access associated public CRL/OCSP in a captive state (rules for this are not included automatically).

**Describe the solution you like**

Update the docs for the functional changes, but revamps other parts as well as needed. Also mention that the automatic rules for CP are now visible in the automatic firewall rules as well and document the RFC8909 feature.

N/A

**Additional context**

N/A

swhite2:
Fixes https://github.com/opnsense/docs/issues/676


Monviech (approved):
LGTM

PR #728, opened 2025-05-16
dnsmasq: Add PTR Records, remove manual setting of router and dns options

Monviech:
Simplify the example a bit as discussed.

### DIFF
--- a/source/manual/dnsmasq.rst
+++ b/source/manual/dnsmasq.rst
@@ -362,11 +362,13 @@ Option                              Value
 **Listen Port**                     ``53``
 ==================================  =======================================================================================================
 
-- Press **Apply**
+- | Press **Apply**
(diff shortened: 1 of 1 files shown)


Monviech:
@fichtner 

Im thinking about setting the search domain to the same value as the domain in the dhcp range via some template magic.
If no dhcp-option for search domain is specified, it will use the default. This will further cut down on setup time.

fichtner:
if that's what ISC DHCP (or perhaps Kea) is doing it can be considered otherwise I wouldn't recommend it to avoid side effects for users switching from one service to the other. Less is explainable, more magic isn't.

fichtner (approved):

Monviech:
It looks like the ``search-domain`` already gets automagically added as option when having a domain set in the ``dhcp-range``

I just tested this in my client network with a range:

- DNS -> automatic
- Router -> automatic
- DNS-Suffix (search domain) -> automatic

This means this manual can be cut down even further.

PR #734, opened 2025-05-21
Update git-backup.rst

7heo:
Bare does not mean empty. From https://git-scm.com/book/en/v2/Git-on-the-Server-Getting-Git-on-a-Server "a new bare repository — a repository that doesn’t contain a working directory".

### DIFF
--- a/source/manual/git-backup.rst
+++ b/source/manual/git-backup.rst
@@ -61,8 +61,9 @@ button in the user interface) is responsible for initialising the empty local re
 
 .. Note::
 
-    One can always change the upstream target, as long as the newly configured one is either "bare" (empty) or containing the
-    exact same content (/change history) as the one used on this firewall.
+    One can always change the upstream target, as long as the newly configured one is either empty (has absolutely no history
+    - otherwise said, be a new one, absolutely untouched), or has the exact same history (otherwise said, be an exact copy, git
+    history included) as the one used on this firewall.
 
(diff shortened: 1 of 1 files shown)


AdSchellevis:
it should be bare, as in `git init --bare my_repo.git`, empty can have a history in which case you can't push.

7heo:
Bare only means that the repository does not have a copy of the files directly on the filesystem. Essentially, it only contains what is normally under the `.git` directory.

I do not see how a bare repository has *any* influence on what can be pushed or not. Bare or not, it is the state of the repository that matters, on which the "bare" status has exactly zero incidence.

An "empty repository", in the sense git means it (as in "warning: You appear to have cloned an empty repository.") does not mean "immediately empty of files", but "empty of history". So in the context of git, "empty" really means what you call "bare". And "bare" means something entirely different.

If you wish to clarify that the remote repository shall have either no history, or the exact same history as the local one, then that is what shall be written.

AdSchellevis:
Changing the text to empty likely won't make it clearer for the user and out of experience causes more noise than it solves, I'm open for suggestions to improve the text, but removing the bare (which doesn't appear to be needed) won't make the issue clearer (init bare or a clean init in this context have the same effect). Keep in mind a lot of people think "empty" equals removing the remote files.

AdSchellevis:
Thanks for making the effort!

PR #746, opened 2025-06-12
dnsmasq: Clear up some points about dhcp reservations

Monviech:


swhite2:
This one is also going to need an update: https://github.com/opnsense/docs/blob/master/source/manual/dhcp.rst?plain=1#L32-L40

PR #765, opened 2025-08-07
Firewall/Aliases - add IPinfo

AdSchellevis:
Add documentation for https://github.com/opnsense/core/issues/7779


source/manual/how-tos/ipinfo_geo_ip.rst
```diff
@@ -0,0 +1,37 @@
...
+Collect the link
+######################
+
+Once you have created an account you'll need to find your license key.
+The easiest process is to login to IPinfo and go to https://ipinfo.io/dashboard/downloads
+
+On the right side of the screen you will notice a big download button, choose the prop format in the tabs above (we need **csv**)
+and use the copy link button next to the download one.
+
+On your clipboard there's no a link which looks like:
```
mrbutcher93, line 24: typo

AdSchellevis: @mrbutcher93 thanks!

meyergru:
What is even worse is something else that I just discovered.... I had problems with the larger aliases because I saw huge latency spikes of ~1500ms, which occured exactly every 60 seconds and caused short VoIP sound outages of 1-2 seconds.

I traced that back to a hidden (root) cron job which calls `/usr/local/bin/flock -n -E 0 -o /tmp/filter_update_tables.lock /usr/local/opnsense/scripts/filter/update_tables.py` every minute. The running time of that job went up to 15 seconds from ~1 second before using IPinfo. This is on a N100 machine.

With the Maxmind database, the latency spikes are much smaller, such that VoIP is not affected.
I could see files in /var/db/aliastables being 10 times larger than with Maxmind,

meyergru:
So people like me will have to buy a CPU that is 10x as fast in order to use the new database. ;-)

Maybe still worth a caveat in the docs, like: 

> For IPinfo, you will need _much_ more alias table space and scale up your CPU accordingly.

PR #766, opened 2025-08-07
update consistent usage of the word instead

scuba-st3v3:
First time contributor! 

- I noticed a few spots that could be updated to the more appropriate usage of the word `instead` vs. `in stead`
- `instead` used 200+ times across various docs already


AdSchellevis:
@scuba-st3v3 better to try to keep this consistent indeed, docs are easy to update

PR #769, opened 2025-08-11
ipsec: clarify NAT situation and document new skip_fw option

swhite2:


AdSchellevis (approved):
@swhite2 looks good, I was doubting if we should mention the fact that `out` is always filtered (which is the cause of the original issue), but it might not increase readability. Let's go with this version.

PR #787, opened 2025-10-22
ipsec: more technical background in tuning considerations section

swhite2:


AdSchellevis (approved):
@swhite2 nice, this helps indeed.

Issue #782, opened 2025-10-14
IP is the same in the document


P0LUCCIA:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [X] I have read the contributing guide lines at https://github.com/opnsense/docs/blob/master/CONTRIBUTING.md
- [X] I have searched the existing issues and I am convinced that mine is new.

**Describe the issue**

https://docs.opnsense.org/manual/how-tos/sslvpn_instance_s2s.html#network-topology

<img width="810" height="588" alt="Image" src="https://github.com/user-attachments/assets/d2abbe4e-def9-40ef-8a74-337adc656543" />

While setting up SSL VPN, I found that the IP address is the same in the protocol.

**Suggestions**

Please change the IP address of Site B Client PC.

PR #806, opened 2025-11-26
sslvpn_instance_s2s.rst: Fix firewall address in network B configuration

yye587:
Fixes issue #782

I presume fwb's IP address is 192.168.8.1/24 based on fwa's IP address (10.0.8.1/24)

### DIFF
--- a/source/manual/how-tos/sslvpn_instance_s2s.rst
+++ b/source/manual/how-tos/sslvpn_instance_s2s.rst
@@ -54,7 +54,7 @@ to peer both firewalls. We will create a tunnel network  :code:`10.1.8.0/24` bet
       }
       network B {
           address = "192.168.8.0/24"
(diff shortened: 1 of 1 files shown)


fichtner:
@Monviech will check, thank you

source/manual/how-tos/sslvpn_instance_s2s.rst
```diff
@@ -54,7 +54,7 @@ to peer both firewalls. We will create a tunnel network  :code:`10.1.8.0/24` bet
       }
       network B {
           address = "192.168.8.0/24"
-          fwb [shape = "cisco.firewall", address="192.168.8.20"];
+          fwb [shape = "cisco.firewall", address="192.168.8.1/24"];
```
Monviech: I would rather give it 192.168.8.2/24 since that matches what the other firewall does (both links having .1).

This one would have .2 on both links that way.

PR #812, opened 2025-11-29
hardware.rst: Raise minimum required RAM from 2 GB to 3 GB

yye587:
https://github.com/opnsense/installer/issues/23


AdSchellevis:
@yye347 thanks, should be 3 nowadays indeed.

PR #828, opened 2026-01-05
Revise MaxMind link creation instructions

jrichman-r7:
Updated instructions for creating a MaxMind link with correct placeholders.


AdSchellevis:
@jrichman-r7 it looks ok, thanks.

PR #838, opened 2026-01-20
Fix typos using codespell

f4ban:
To test [codespell](https://github.com/codespell-project/codespell) for other projects, I did a quick run here to see how it works. I was surprised by how many typos there were.

I went through them manually to make sure the fixes are correct.

Happy reviewing :D

### DIFF
--- a/source/manual/vpnet.rst
+++ b/source/manual/vpnet.rst
@@ -720,7 +720,7 @@ The following functions are available in the menu (as of OPNsense 23.7):
 
 * Connection Status
 
-  * Show tunnel statuses
+  * Show tunnel status
 
 * Log File
 


source/manual/how-tos/ntopng.rst
```diff
@@ -23,7 +23,7 @@ General Settings
 :Interfaces:
     Here you set the interfaces ntopng should listen on. If you don't select any interface
     it listens to the first in the system, e.g. em0, but you can change the interfaces 
-    within ntopng's UI on demand; while setting an explicit interface you wont get any 
+    within ntopng's UI on demand; while setting an explicit interface you won't get any
```
fichtner:
```suggestion
    within ntopng's UI on demand; while setting an explicit interface you will not get any 
```

source/manual/dnsmasq.rst
```diff
@@ -492,7 +492,7 @@ In our example, we configure query forwarding for 2 networks:
 .. Note::
 
     ``.internal`` is the IANA and ICANN approved TLD (Top Level Domain) for internal use. If you instead own a TLD, e.g., ``example.com``, you could create a zone
-    thats not used on the internet, e.g., ``lan.internal.example.com``.
+    that's not used on the internet, e.g., ``lan.internal.example.com``.
```
fichtner:
```suggestion
    that is not used on the internet, e.g., ``lan.internal.example.com``.
```

source/manual/vpnet.rst
```diff
@@ -720,7 +720,7 @@ The following functions are available in the menu (as of OPNsense 23.7):
 
 * Connection Status
 
-  * Show tunnel statusses
+  * Show tunnel statuses
```
fichtner:
```suggestion
  * Show tunnel status
```

fichtner (approved):
small changes to table layouts, "statuses" to "status" as we very rarely use it in the GUI, avoid a contraction in a changed line while we are here (it is a technical document after all) and once instance where "addresss" was "addresses" instead.

Overall very good change. Relatively normal for a human-made documentation mostly written by non-native speakers. Thank you.

PR #847, opened 2026-02-06
Tayga: creating a normalization rule is no longer required

maurice-w:
opnsense/core@630ab19 added a global "scrub in all" rule. Normalization rules for individual interfaces are no longer needed.


fichtner:
Merged, thanks!

PR #868, opened 2026-04-14
thefork: fix links 

sopex:
I know https://www.deciso.com/2015-january-2nd-deciso-launches-opnsense-a-new-open-source-firewall-initiative/ exists, but it lacks the vibe of the era that a page like thefork would benefit.

But it could also be used.


AdSchellevis:
@sopex thanks for fixing the links, I see what you mean with the link to the archive page, the only downside is that it's extremely slow, but for historic context that's probably not very relevant anyway.

sopex:
It really is :)

PR #889, opened 2026-06-20
Add note for os-cloudflared about ICMP proxy not implemented on FreeBSD

insanityinside:
Note regarding https://github.com/opnsense/plugins/issues/5511 - ICMP proxy is not implemented by cloudflared daemon on FreeBSD, so warning will always fire until it's fixed. Didn't filter the warning from the log in case it does get fixed upstream at some point, but worth warning users about the noise.


Monviech:
Thanks :)

PR #890, opened 2026-06-29
How-To LAN bridge: removal of port isolation

### LINKED ISSUE #10468: OPNsense 26.4.1 Business Edition: Bridge interface blocks L2-forward for IP (while ARP is working)
**Important notices**

- [X] I have read the contributing guidelines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [X] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the bug**

I am migrating an existing bridge-based VLAN setup from an older OPNsense installation.

The setup follows the documentation for a layer-2-bridge as described here: https://docs.opnsense.org/manual/how-tos/lan_bridge.html

Given a bridge consisting of two VLAN interfaces:

- vlan0.12.30 (parent: lagg0(igc1,igc2))
- vlan1.0.30 (parent: ax0)

The bridge itself has an IPv4 address for routing.

The devices behind both vlans can communicate with all other subnets and vice versa (Layer 3, routing involved). But the devices behind vlan0.12.30 cannot communicate with the devices in the same subnet behind vlan1.0.30 and vice vesa (Layer 2, bridging use case).

**The surprising part is that the bridge behaves like a transparent Layer-2 bridge for ARP, but requires an explicit Layer-3 pass rule for forwarding IP traffic between bridge members. This is unexpected because no routing is involved.**

**To Reproduce**

Bridge filtering sysctls are set:

net.link.bridge.pfil_bridge=1
net.link.bridge.pfil_member=0
net.link.bridge.pfil_local_phys=0

1. Setup a bridge as described
2. Connect one client device to vlan0.12.30
3. Connect another client device to vlan1.0.30
4. arping from device 1 to device 2 --> passed
5. ping from device 1 to device 2 --> blocked (default deny / state violation)

**Expected behavior**

Following the given documentation, I expected a layer-2-bridging, not only for ARP but also for IP. IIRC this was the behavior in the older OPNSense setup, what I unfortunately cannot test anymore.

_Workaround_  

Adding the following firewall rule on the BRIDGE interface immediately restores forwarding:

- Action: Pass
- Interface: bridge30
- Source: bridge30 net
- Destination: bridge30 net
- Resulting pf rule: pass in log quick on bridge30 inet from (bridge30:network) to (bridge30:network) keep state

**Relevant log files**

Packet captures show:

- lagg0: ICMP Echo Request visible
- vlan0.12.30: ICMP Echo Request visible
- bridge30: ICMP Echo Request visible
- vlan1.0.30: NO ICMP packet
- ax0: NO ICMP packet

**Additional context**

Has bridge filtering semantics changed recently?

If this is intentional, I think the documentation (e.g. https://docs.opnsense.org/manual/how-tos/lan_bridge.html) should explicitly mention that bridge-local traffic must be permitted by firewall rules on the bridge interface when pfil_bridge=1.

If it is not intentional, this may be a regression.

**Environment**

Software version used and hardware type if relevant, e.g.:

OPNsense 26.4.1 Business Edition
FreeBSD 14.3-RELEASE-p15
DEC740

Sumpfdotter:
Added a new "Port Isolation" section ("Step 7") to the LAN bridge how-to.

Fixes: https://github.com/opnsense/core/issues/10468

The original how-to leaves users with ARP port isolation disabled but IP port isolation still enabled because traffic between bridge member interfaces is filtered by the firewall by default.

The new section explains this behavior and shows how to allow IP traffic between bridge member interfaces when port isolation is not desired.


source/manual/how-tos/lan_bridge.rst
```diff
@@ -81,6 +85,20 @@ Select the tunable net.link.bridge.pfil_bridge and set the value to 1
 .. image:: images/lan_bridge_7.png
 	:width: 100%
 
+**Step Seven**
+-----------------    
+
+OPT1, OPT2, and OPT3 are now configured as a single broadcast domain. However, IP traffic between bridge member
```
fichtner, line 91: I'd suggest mentioning `net.link.bridge.pfil_onlyip` here for completeness as it influences the claim of "single broadcast domain".

Sumpfdotter: I had the same thought at first. Then I reviewed the document again while looking at all bridge-related tunables:

net.link.bridge.ipfw: 0
net.link.bridge.member_ifaddrs: 1
net.link.bridge.log_mac_flap: 1
net.link.bridge.allow_llz_overlap: 0
net.link.bridge.inherit_mac: 0
net.link.bridge.log_stp: 0
net.link.bridge.pfil_local_phys: 0
**net.link.bridge.pfil_member: 0**
net.link.bridge.ipfw_arp: 0
**net.link.bridge.pfil_bridge: 1**
net.link.bridge.pfil_onlyip: 0

It turned out that the document only mentions tunables that need to be **changed from their default values**.

I want to either explain all of the bridge tunables or only document the ones that deviate from the defaults. Since the current how-to intentionally contains very little low-level technical detail, I chose not to clutter it with a complete list of tunables.

fichtner: Defaults may change and people have different focus to the document. It's ok for me to leave as suggested but I wanted to illustrate how you think to close the gap in the document but also introduce another point of future contention from someone else. Not everyone who is dissatisfied with the documentation will raise a ticket or mention it somewhere for us to see.

PR #892, opened 2026-07-21
development/api: Firewall - drop removed savepoint/rollback actions

### LINKED ISSUE #891: Firewall API reference documents removed savepoint/rollback actions (removed in core #10411)
## Problem

The Firewall API reference at
https://docs.opnsense.org/development/api/core/firewall.html still
documents a savepoint / rollback mechanism that no longer exists in core.

The page describes:
- a `savepoint()` action,
- `apply(savepoint)` taking a rollback revision,
- a `cancelRollback(sp)` action with a 60-second auto-revert timer,
- a `revert($revision)` action,

and ships a `savepoint_example.py` built around them.

## Why it's stale

These actions were removed from core in opnsense/core#10411
(commit 17b84612eb, merged 2026-06-18, first released in **26.7**),
tracked by opnsense/core#10235 ("Deprecate and remove savepoint actions",
labeled *cleanup*, described as "a remainder of the os-firewall plugin").

In current core, `FilterBaseController::applyAction()` takes no arguments
and simply runs `configd filter reload skip_alias`; the `savepoint`,
`cancelRollback` and `revert` actions, the `rollback_timer.php` /
`rollback_cancel.php` scripts and their configd actions are all gone.
`filter_configure_sync()` performs a plain reload — there is no
config-revision rollback in the apply path anymore.

The auto-generated endpoint table on that page also still lists
`apply,$rollback_revision=null`, `cancel_rollback`, `revert` and
`savepoint`; it appears to have last been regenerated against 26.1.

## Open question

The core issue frames this as permanent cleanup of an os-firewall plugin
remainder, and I could not find any issue, PR or discussion indicating the
feature is meant to return. That said, the firewall is actively migrating
from legacy PHP to MVC, and a maintainer noted on opnsense/core#10411 a
possible future generic MVC facility for change tracking / the apply
button — so I cannot rule out that the docs were intentionally left in
place. If reintroduction is planned, please close this; otherwise a PR
removing the stale content follows.

chrstnth:
The Firewall API reference documented a savepoint / rollback mechanism
(`savepoint()`, `apply(savepoint)`, `cancelRollback(sp)`, `revert()`) that
was removed from core in opnsense/core#10411 (released in 26.7, tracked by
opnsense/core#10235).

This updates the reference to the current apply-only flow:
- rewrite the Concept section and blockdiag to a plain `apply()` step
- drop the rollback-related Tip and Note
- replace the savepoint example with a plain apply example
  (`firewall.apply.py`, replacing `firewall.savepoint.py`)
- clean the `filter_base` endpoint table (`apply` without
  `$rollback_revision`, remove `cancel_rollback` / `revert` / `savepoint`)

The `filter_base` table was edited in place rather than regenerated in
full, to keep this change scoped to the savepoint removal; the rest of the
firewall endpoint tables carry unrelated drift that a full
`collect_api_endpoints.py` refresh would pick up.

closes #891

### DIFF
--- a/source/development/api/core/firewall.apply.py
+++ b/source/development/api/core/firewall.apply.py
@@ -22,17 +22,14 @@ if r.status_code == 200:
     response = json.loads(r.text)
     if len(response['rows']) > 0:
         rule_uuid = response['rows'][0]['uuid']
-        r = requests.post("%s/api/firewall/filter/savepoint" % remote_uri, auth=(api_key, api_secret), verify=False)
-        if r.status_code == 200:
-            sp_response = json.loads(r.text)
-            # disable rule
-            r = requests.post("%s/api/firewall/filter/toggleRule/%s/0" % (remote_uri, rule_uuid),
-                              auth=(api_key, api_secret), verify=False
-            )
-            # apply changes, revert to sp_response['revision'] after 60 seconds
-            r = requests.post("%s/api/firewall/filter/apply/%s" % (remote_uri, sp_response['revision']),
-                              auth=(api_key, api_secret), verify=False
-            )
-            print("revert to revision %s in 60 seconds (%s changed)" % (sp_response['revision'], rule_uuid))
+        # disable rule
+        r = requests.post("%s/api/firewall/filter/toggleRule/%s/0" % (remote_uri, rule_uuid),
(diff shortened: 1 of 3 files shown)


AdSchellevis:
@chrstnth thanks, we forgot about the docs here.

PR #906, opened 2026-09-03
vpn: Remove the legacy IPsec guides and adjust the vpnet manual to reflect mostly connections

Monviech:
23.1 was a long time ago, time to let go.


Monviech:
@fichtner moving this to you for now

Monviech:
I think its sensible to remove these, most of these guides are outdated and for very old e.g. opnsense or client software versions. With the ikev1 removal on the horizon there is even less of a reason to keep them anyway.
https://github.com/opnsense/core/issues/10874

fichtner:
I wasn't sure about good timing, but I don't disagree here. Thanks!

PR #908, opened 2026-09-06
Add info_link for static text as clickable links

sopex:
@AdSchellevis 

https://github.com/opnsense/core/pull/10848


AdSchellevis:
@sopex thanks a lot!

